PR #17, opened 2017-03-17
New color palettes [palettes-dev]

tzanio:
Added a few new interesting color palettes from

http://www.ncl.ucar.edu/Document/Graphics/color_table_gallery.shtml

and finally added the ability to go backwards to the previous palette (key 'P', printing now is 'Ctrl+P').

Total number of palette options now is 35. Here they are (left-to-right, top-to-bottom) in 2D:

palettes-2d

Here is an example of Laplace on 3D surface:

palettes-3d


tzanio:
Hey, it's my day off :)

PR #39, opened 2017-11-10
Add capability to show element and vertex numberings in 2D [numberings-dev]

rw-anderson:
Cycle through none, element, and vertex numberings with the 'n' key.

A few issues:
* There are some tricky issues with occlusion depending on what you are looking at.  It may be best not to fight this, but just to turn off the elements to work around this when needed.
* In parallel, vertex numberings overwrite one another.  Shrinking domains with F11 is a workaround for this.  The labels of both elements and vertices should follow any shrinking transformations.
* Not implemented in 3D.
* Text drawing gets quite slow for large numbers of entities.  I put in a cutoff above 1000 entities to just disable the feature.  I figure since this feature is probably most useful in a debugging context, this isn't a deal breaking compromise.  It can be increased if you are patient.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,6 +12,8 @@
 Development version, not released yet
 =====================================
 
+- Added capability to show element and vertex numbering in 2D (key 'n').
+
 - Added support for reading mesh and solution from the same file.
 
 - Added a CMake build system.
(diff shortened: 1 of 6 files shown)


tzanio:
@rw-anderson, feel free to merge the PR and delete the numberings-dev branch if you are OK with the changes.

PR #80, opened 2019-03-08
Moves palettes.hpp to palettes.cpp, uses header instead of forward declarations

tomstitt:
* Moves most of *palettes.hpp* into *palettes.cpp* so that *palettes.hpp* can be included in more than one file without duplicate symbols

* Remove forward declarations in *vsdata.cpp* and *threads.cpp* and includes *palettes.hpp* instead

I was merging master into the modern opengl branch but ran into conflicts with palettes, I'm making the changes here so they're easier to follow.


lib/openglvis.cpp
```diff
@@ -66,10 +66,10 @@ void Camera::GLMultRotMatrix()
    double mat[16] =
    {
       -left[0], up[0], -dir[0], 0.0,
-         -left[1], up[1], -dir[1], 0.0,
```
tomstitt: Interesting, when I run `make style` it puts these indents in... what version of *astyle* are you using?

tzanio: Version 2.05.1, see https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md#developer-guidelines :)

PR #96, opened 2020-01-07
Display an element ordering curve using the o key [element-ordering-curve-dev]

rw-anderson:
This is maybe a little specialized, but since I wrote it while working on element ordering, I thought I would push it up.  Use the 'o' key to display an element ordering curve on 2D meshes.

Examples:

With arrows (first 'o' press):
star-mesh-ordering-curve-arrows

Without arrows (second 'o' press):
star-mesh-ordering-curve-noarrows

AMR mesh:
amr-mesh-ordering-curve

@jakubcerveny 
@tzanio 

| [#96](https://github.com/glvis/glvis/pull/96) | @rw-anderson | @tzanio | @jakubcerveny + @acfisher | 04/25/20 | 05/19/20 | 05/19/20 | |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,7 +26,7 @@ Version 3.4.1 (development)
 
 - Added a key for setting the bounding box from the terminal (Shift+F7).
 
-- Add support to visualize an element ordering curve (key 'ctrl+o').
+- Add support to visualize the element ordering curve with 'Ctrl+o'.
 
 Version 3.4, released on May 29, 2018
 =====================================
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,7 +26,7 @@ Version 3.4.1 (development)
 
 - Added a key for setting the bounding box from the terminal (Shift+F7).
 
-- Add support to visualize an element ordering curve in 2D (key 'o').
+- Add support to visualize an element ordering curve (key 'ctrl+o').
 
 Version 3.4, released on May 29, 2018
 =====================================
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,6 +26,7 @@ Version 3.4.1 (development)
 
 - Added a key for setting the bounding box from the terminal (Shift+F7).
 
+- Add support to visualize an element ordering curve in 2D (key 'o').
 
 Version 3.4, released on May 29, 2018
 =====================================
--- a/README
+++ b/README
@@ -141,6 +141,7 @@ arrow keys        - Manual rotation
 1,2,3,4,5,6,7,8,9 - Manual rotation along coordinate axes
 Ctrl+arrow keys   - Translate the viewpoint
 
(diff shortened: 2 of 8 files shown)


tzanio:
Very cool! That's what I was referring to in https://github.com/mfem/mfem/pull/1166#discussion_r362575376

jakubcerveny:
Nice! This is useful. I'm just not sure about the `o` key, isn't that for subdivision? Also, why not make it work in 3D too?

rw-anderson:
I think the i key is for subdivision.  The o key was apparently unused.

Implementing in 3D should be doable, I just wasn’t sure if the result would be useful.

acfisher:
I like it.  I'm sure you guys will sort out the interface issue.

tzanio:
> Render the curve without the arrowheads (these will quickly clutter the image when the mesh gets finer or when you switch to 3D). Instead, color the curve by the palette, see images here: mfem/mfem#697. This will work relatively OK in 3D too (when Z-buffered).

I think this is a good idea and will be nice to implement in this PR.

> Fix the inconsistency in GlVis by assigning the o/O keys to subdivision both in 2D and in 3D. Keep the i/I keys for 3D clipping.

In an informal poll at the last MFEM PR meeting (surprisingly to me) everyone supported this. Unless @v-dobrev has strong objections, I'd say go for it.

> Toggle the ordering curve with some less prominent key (Ctrl+O could work but maybe there is something better) in both 2D and 3D.

Agreed

v-dobrev:
> Unless @v-dobrev has strong objections, I'd say go for it.

I have no objections. This specific update should be mentioned in the CHANGELOG, and reflected in the documentation (all the versions of the `h` key, as well as the README file).

tzanio:
@rw-anderson, what do you think about [Jakub's suggestions](https://github.com/GLVis/glvis/pull/96#pullrequestreview-400964688)?

rw-anderson:
I like all of the suggestions.  I am not immediately sure how to do the coloring but will take a closer look at it.

rw-anderson:
noarrow
Color from palette, no arrows, by default

witharrows
Since the arrowheads are sized by segment length, they help bring out the locations of big jumps, so I left them in as an option to toggle through.

rw-anderson:
3d-ordering-curve

3D ordering curve.

rw-anderson:
Making the lines a little thicker makes them distinguishable from the mesh.
mobius-3d

rw-anderson:
I implemented 3d, coloring, and bound the ordering curve to ctrl-o in 2D and 3D.

Fixing the i vs o inconsistency in 2D and 3D is kind of a separate issue so I think I'm going to punt on that for right now and file it as an issue to be addressed in a separate PR.

jakubcerveny:
I think it's great! Thanks!

Tried it on a couple of meshes, it seems to work nicely.

ordering

PR #109, opened 2020-10-09
Don't open all files at once when reading parallel meshes/solutions.

### LINKED ISSUE #55: Too many open files when reading parallel mesh & solution
After some recent changes in GLVis (meshes and solutions in the same file), it is no longer able to open a larger parallel solution (like 4096 CPUs). The error you get is something like

```Can not open mesh file: sedov_500_mesh.001013!```

while the file clearly exists and can be opened. It turns out GLVis is trying to open all of the files at once and the default limit on Ubuntu is ~1000 concurrently open files per process. The workaround is to increase the limit with something like `ulimit -n 10000`, but I think it shouldn't be necessary to keep all the files open and that we should eventually fix this.

jakubcerveny:
Fix #55.

| [#109](https://github.com/glvis/glvis/pull/109) | @jakubcerveny | @tzanio | @tzanio + @dylan-copeland | 10/09/20 | 10/09/20 | 10/09/20 | |

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -39,6 +39,9 @@ Version 3.4.1 (development)
 
 - Add support to visualize the element ordering curve with 'Ctrl+o'.
 
+- Improved the opening of parallel meshes/solutions.
+
+
 Version 3.4, released on May 29, 2018
 =====================================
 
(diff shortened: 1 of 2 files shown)


tzanio:
@dylan-copeland, can you approve?

tzanio (approved):
Thanks @jakubcerveny !

dylan-copeland (approved):
Thanks @jakubcerveny ! Note that my review of the code was superficial, as I am not familiar with the glvis code. I tested on a very large number of files, as well as ex1p and ex25p.

tzanio:
@kanye-quest, do you think we should merge this fix in `master` and then in your PR, or merge in your PR directly, or something else?

PR #127, opened 2020-12-09
Fix for numpad keys issuing duplicate commands

kanye-quest:


kanye-quest:
@v-dobrev, can you see if this fixes the issue with the keypad commands?

v-dobrev (approved):
Seems to work as expected now.

PR #135, opened 2020-12-21
Shared stream reader

tomstitt:
Moves mostly duplicate `ReadStream`, `SetMeshSolution`, and `Extrude1DMeshAndSolution` from _glvis.cpp_ and _lib/aux_js.cpp_ to _lib/stream_reader.{hc}pp_

Adds `ASTYLE_BIN` option to the makefile

Tested on OS X and Emscripten with saved stream data.


lib/stream_reader.hpp
```diff
@@ -0,0 +1,37 @@
...
+   std::string keys;
+   mfem::Mesh *mesh{nullptr};
+   mfem::GridFunction *grid_f{nullptr};
+   int is_gf{0};
+   bool fix_elem_orient{false};
+   bool save_coloring{false};
+};
+extern StreamState stream_state;
+
+void Extrude1DMeshAndSolution(mfem::Mesh **mesh_p,
```
kanye-quest, line 30: Could these methods be made members of `StreamState`?

glvis.cpp
```diff
@@ -106,235 +111,6 @@ GridFunction *ProjectVectorFEGridFunction(GridFunction*);
 void Extrude1DMeshAndSolution(Mesh **mesh_p, GridFunction **grid_f_p,
```
kanye-quest, line 111: I think this declaration can be removed.

glvis.cpp
```diff
@@ -56,11 +57,15 @@ string      extra_caption;
 
 // Global variables
 int input = 1;
-Mesh *mesh = NULL;
-Vector sol, solu, solv, solw, normals;
-GridFunction *grid_f = NULL;
-int is_gf = 0;
-string keys;
+Mesh *&mesh = stream_state.mesh;
```
kanye-quest, line 60: Are there any methods left in glvis.cpp that rely on these variables, or could we just pass around the `stream_state` object?

lib/stream_reader.hpp
```diff
@@ -0,0 +1,37 @@
...
+{
+   mfem::Vector sol, solu, solv, solw, normals;
+   std::string keys;
+   mfem::Mesh *mesh{nullptr};
+   mfem::GridFunction *grid_f{nullptr};
+   int is_gf{0};
+   bool fix_elem_orient{false};
+   bool save_coloring{false};
+};
+extern StreamState stream_state;
```
kanye-quest, line 28: I'd personally prefer this variable be created in glvis.cpp as it's needed, so we don't accidentally introduce any extra dependencies on global state in the future.

kanye-quest (commented):
LGTM, just a few comments/suggestions I had.

glvis.cpp
tomstitt: Yeah, at least `GLVisInitVis`, `GLVisStartVis`, `ExecuteScriptCommand`, and `PlayScript`.  We could switch to using an object but that's a bigger diff and I think it would be better for a followup PR.

lib/stream_reader.hpp
tomstitt: They could but I think that would be a larger refactor that makes more sense for a followup PR. We'd need to refactor _glvis.cpp_ a bit for it to not rely on globals.

kanye-quest: Yeah, that's fine with me.

makefile
```diff
@@ -212,7 +212,7 @@ Ccc  = $(strip $(CC) $(CFLAGS) $(GL_OPTS))
 ALL_SOURCE_FILES = \
  lib/gl/renderer.cpp lib/gl/renderer_core.cpp lib/gl/renderer_ff.cpp \
  lib/gl/types.cpp lib/aux_js.cpp lib/aux_vis.cpp lib/font.cpp lib/gl2ps.c \
- lib/material.cpp lib/openglvis.cpp lib/palettes.cpp lib/sdl.cpp \
+ lib/material.cpp lib/openglvis.cpp lib/palettes.cpp lib/sdl.cpp lib/stream_reader.cpp \
```
kanye-quest, line 215: These should be added to the cmake build as well.

tomstitt: Thank you! Fixed.

tzanio (approved):

kanye-quest (approved):

PR #139, opened 2020-12-30
GetHelpString

tomstitt:
Adds `SdlWindow::getHelpString` and exports as JavaScript function.

Emscripten builds cause `stdout` to use `console.log` but Emscripten also inlines the definition of `console.log` (at module load time?) so we can't dynamically change `console.log` to temporarily write the output from `sendKey('h')` to a string.

### DIFF
--- a/lib/sdl.cpp
+++ b/lib/sdl.cpp
@@ -831,13 +831,3 @@ void SdlWindow::swapBuffer()
 {
    SDL_GL_SwapWindow(handle->hwnd);
 }
-
-std::string SdlWindow::getHelpString()
-{
-   std::ostringstream oss;
-   auto * fb = std::cout.rdbuf();
-   std::cout.rdbuf(oss.rdbuf());
-   onKeyDown['h'](0);
-   std::cout.rdbuf(fb);
-   return oss.str();
-}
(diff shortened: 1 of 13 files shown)


tzanio (approved):

lib/sdl.cpp
```diff
@@ -831,3 +831,13 @@ void SdlWindow::swapBuffer()
...
 }
+
+std::string SdlWindow::getHelpString()
+{
+   std::ostringstream oss;
+   auto * fb = std::cout.rdbuf();
+   std::cout.rdbuf(oss.rdbuf());
+   onKeyDown['h'](0);
+   std::cout.rdbuf(fb);
+   return oss.str();
```
tzanio, line 842: This works, but I wonder it it will be better to expose the static functions
`SolutionKeyHPressed` in `lib/vssolution.cpp`, `Solution3dKeyHPressed` in `lib/vssolution3d.cpp, and `VectorKeyHPressed` in `lib/vsvector.cpp` and `lib/vsvector3d.cpp`.

@v-dobrev and @tomstitt -- what do you think?

tomstitt: Here's another option: https://github.com/GLVis/glvis/pull/142

tzanio: I'd like to hear from @v-dobrev, but I personally like the #142 option more

tzanio (approved):
Looks good to me!

kanye-quest (approved):

PR #142, opened 2021-01-04
GetHelpString v2

tomstitt:
Register a help function with an `std::ostream &` argument instead of hijacking `cout`

### DIFF
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -46,8 +46,9 @@ int isfinite(double x)
 
 // Definitions of some more keys
 
-static void SolutionHelpFunction(std::ostream & os)
+std::string VisualizationSceneSolution::GetHelpString() const
 {
+   std::stringstream os;
    os << endl
       << "+------------------------------------+" << endl
       << "| Keys                               |" << endl
@@ -117,6 +118,7 @@ static void SolutionHelpFunction(std::ostream & os)
       << "| right + Ctrl  - Object scaling     |" << endl
       << "| left  + Ctrl + Shift - z-Spinning  |" << endl
       << "+------------------------------------+" << endl;
+   return os.str();
 }
 
 static void KeyF8Pressed()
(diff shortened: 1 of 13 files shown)


lib/vssolution.cpp
```diff
@@ -46,77 +46,77 @@ int isfinite(double x)
...
-        << "| right  btn    - Scaling            |" << endl
-        << "| left  + Alt   - Tilt               |" << endl
-        << "| left  + Shift - Spinning           |" << endl
-        << "| right + Shift - Change light pos.  |" << endl
-        << "| left  + Ctrl  - Spherical rotation |" << endl
-        << "| middle+ Ctrl  - Object translation |" << endl
-        << "| right + Ctrl  - Object scaling     |" << endl
-        << "| left  + Ctrl + Shift - z-Spinning  |" << endl
-        << "+------------------------------------+" << endl;
+static void SolutionHelpFunction(std::ostream & os)
```
kanye-quest, line 49: We could go one step further, and make these member virtual functions on the VisualizationSceneSolution classes that return an `std::string`. Then we could avoid having to go through the SdlWindow to get the help string.

tzanio: I agree.

I don't remember anymore why these functions were declared to be static. Maybe there was a reason for it with the old X11 code, but if it is not necessary anymore having them as member functions will be much cleaner.

@kanye-quest do you want to make the change? (I can make it too)

tomstitt: I think the issue with this in the current code would be that `SdlWindow::onKeyDown` uses function pointers so a method wouldn't work, but it could be changed to use `std::function` so we could use a lambda with the captured `VisualizationSceneSolution` instead

kanye-quest:
@tomstitt and @tzanio, I added some changes to handle the help string with virtual functions. Feel free to fix/revert as needed :)

tzanio (approved):
I like this version, thanks @tomstitt and @kanye-quest

tomstitt:
> @tomstitt and @tzanio, I added some changes to handle the help string with virtual functions. Feel free to fix/revert as needed :)

Looks great! Thanks @kanye-quest

PR #147, opened 2021-01-18
Parse the valuerange command in JS streams

tzanio:
This is useful for some of the saved stream in the GLVis website, e.g. 

https://github.com/GLVis/web/blob/gh-pages/data/mfem-logo.saved

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -62,12 +62,18 @@ bool startVisualization(const std::string input, const std::string data_type,
    GetAppWindow()->getRenderer().setAntialiasing(0);
 
    std::string line;
+   double minv = 0.0, maxv = 0.0;
    while (ss >> line)
    {
       if (line == "keys")
       {
          std::cout << "parsing 'keys'" << std::endl;
          ss >> stream_state.keys;
+      }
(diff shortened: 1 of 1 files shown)


lib/aux_js.cpp
```diff
@@ -62,13 +62,19 @@ bool startVisualization(const std::string input, const std::string data_type,
...
       if (line == "keys")
       {
          std::cout << "parsing 'keys'" << std::endl;
          ss >> stream_state.keys;
       }
+      else if (line == "valuerange")
+      {
+         std::cout << "parsing 'valuerange'" << std::endl;
+         ss >> minv >> maxv;
+      }
```
tomstitt, line 77: We should (but not now) think about making a shared parser with *glvis.cpp* for the end-of-stream messages

tomstitt (approved):

lib/aux_js.cpp
tzanio: I agree, I thought about this too, but didn't have the time to do it carefully.

(plus "valuerange" was the only thing I needed for the additional examples)

kanye-quest (approved):

PR #149, opened 2021-02-05
[WIP] Multithreaded refactoring

kanye-quest:
## Summary
Refactoring of GLVis to make it more amenable to opening multiple solutions within a single process. This avoid the need to `fork` in client-server mode, which may cause some issues with the native window framework on Mac OS X.

Related to #141.

## Completed
- Consolidates global state in `aux_vis.cpp` into `GLVisWindow` type, which handles creation of associated `VisualizationScene` and `GLVisCommand` objects
- Move `StreamReader` functions into the object itself
- Modularizes palette code into `PaletteState` object
- Replaces pthreads in `GLVisCommand` with C++ standard library threads
- Consolidate some shared code between `threads.cpp`, `glvis.cpp`, `aux_js.cpp`
  - See `StreamState::NewMeshAndSolution`, `StreamState::CreateVisualizationScene`
## To be completed
- [x] Make `VisualizationScene` key event handlers rely on specific instance of `GLVisWindow`/`VisualizationScene` instead of a global pointer
- [ ] Modify `SdlWindow` to support multiple windows with single event loop
- [x] Handle `GLVisFont` as a singleton, to avoid multiple initializations of Freetype
- [ ] Replace fork() in client-server mode with opening a new `GLVisWindow` on a new thread


kanye-quest:
Closing in favor of #170 and #174.

PR #151, opened 2021-02-10
Add key processing JavaScript callbacks

tomstitt:


lib/aux_js.cpp
```diff
@@ -334,6 +334,20 @@ void setKeyboardListeningElementId(const std::string & id)
...
+  CallKeySequence(keys.c_str());
+}
+
+void processKey(char sym, bool ctrl=false, bool shift=false, bool alt=false)
+{
+  Uint16 mod = 0;
+  mod |= ctrl ? KMOD_CTRL : 0;
+  mod |= shift ? KMOD_SHIFT : 0;
+  mod |= alt ? KMOD_ALT : 0;
+  GetAppWindow()->callKeyDown(sym, mod);
```
kanye-quest, line 348: I think this should be `signalKeyDown(sym, mod)` instead of `callKeyDown(sym, mod)`; IIRC `callKeyDown()` and `CallKeySequence()` only are needed as a special case for scripts and GLVisCommand, since they might submit a screenshot command that needs to be handled synchronously.

lib/aux_js.cpp
```diff
@@ -334,6 +334,20 @@ void setKeyboardListeningElementId(const std::string & id)
    SDL_SetHint(SDL_HINT_EMSCRIPTEN_KEYBOARD_ELEMENT, id.c_str());
 }
 
+void processKeys(const std::string & keys)
+{
+  CallKeySequence(keys.c_str());
```
kanye-quest, line 339: Same here with `SendKeySequence()` instead of `CallKeySequence()`?

tomstitt: I was following something @tzanio added in https://github.com/GLVis/glvis/commit/fe28c06c8eab4b899a5c0e33cd5f68c882aba035 , and I think we want `CallKeySequence` so that the keys are processed in a batch, otherwise you will visually see each applied in sequence instead of just the result.

lib/aux_js.cpp
tomstitt: Oh, interesting, that's a good question! I think we'll need to demo both in a live app to see. There are some changes that mean the literal keyboard presses are going through this interface now and I'm not sure what's going to be better

lib/aux_js.cpp
tzanio: I think this ☝️ is right. For the web visualization `CallKeySequence` is better

lib/aux_js.cpp
tzanio: I don't have a strong opinion about this. If it is just about the JS code, I don't think we need to worry about screenshots.

tzanio (approved):
Seems to work great in `glvis-js`, thanks @tomstitt !

lib/aux_js.cpp
kanye-quest: Based on the other discussion, this should be okay.

kanye-quest (approved):
LGTM 👍

PR #153, opened 2021-02-16
StreamState refactor

kanye-quest:
- Make stream_reader.hpp functions members of `StreamState`
- Consolidate code to update mesh and solution into `StreamState::NewMeshAndSolution`, used by threads.cpp, aux_js.cpp, and glvis.cpp
- Where possible, remove references to global `StreamState stream_state` instance

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -21,6 +21,7 @@ std::string extra_caption; // used in extern context
 mfem::GeometryRefiner GLVisGeometryRefiner; // used in extern context
 
 static VisualizationSceneScalarData * vs = nullptr;
+StreamState stream_state;
 
 namespace js
 {
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -56,7 +56,7 @@ bool startVisualization(const std::string input, const std::string data_type,
    std::stringstream ss(input);
 
    // 0 - scalar data, 1 - vector data, 2 - mesh only, (-1) - unknown
-   const int field_type = ReadStream(ss, data_type);
+   const int field_type = stream_state.ReadStream(ss, data_type);
 
    // reset antialiasing
    GetAppWindow()->getRenderer().setAntialiasing(0);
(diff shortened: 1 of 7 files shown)


lib/stream_reader.cpp
```diff
@@ -42,35 +38,31 @@ void Extrude1DMeshAndSolution(Mesh **mesh_p, GridFunction **grid_f_p,
...
       GridFunction *grid_f_2d =
-         Extrude1DGridFunction(mesh, mesh2d, *grid_f_p, 1);
+         Extrude1DGridFunction(mesh.get(), mesh2d, grid_f.get(), 1);
 
-      delete *grid_f_p;
-      *grid_f_p = grid_f_2d;
+      grid_f.reset(grid_f_2d);
    }
-   if (sol && sol->Size() == mesh->GetNV())
+   else if (sol.Size() == mesh->GetNV())
```
tomstitt, line 50: Why was this changed to an `else if`? Did I copy it wrong before?

lib/aux_js.cpp
```diff
@@ -244,64 +247,23 @@ int updateVisualization(std::string data_type, std::string stream)
...
-            vss->NewMeshAndSolution(new_m, &stream_state.sol, new_g);
-         }
-         else
-         {
-            new_g = ProjectVectorFEGridFunction(new_g);
-            VisualizationSceneVector3d *vss =
-               dynamic_cast<VisualizationSceneVector3d *>(vs);
-            vss->NewMeshAndSolution(new_m, new_g);
-         }
-      }
```
tomstitt, line 286: much cleaner! nice!

tzanio (approved):
I have not checked all the refactoring changes super-carefully, but it didn't notice any issues and the code appears to work as before.

tomstitt (approved):
I looked over everything briefly and the changes look good to me. I tested with glvis-js and everything I tested looks like it's working (loading files/updating with streams)

PR #156, opened 2021-02-18
Pinch to zoom

tomstitt:
- Adds event hander for `SDL_MULTIGESTURE`; pinch and rotate
   - Adds pinch handler that zooms
   - No rotate handler currently
   - Mouse Events (single finger touch events) are disabled when the number of detected fingers is > 1
- `void processKey(char sym, ...` -> void processKey(int sym, ...` for JS interface so that high-valued functions keys work

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -339,7 +339,7 @@ void processKeys(const std::string & keys)
    CallKeySequence(keys.c_str());
 }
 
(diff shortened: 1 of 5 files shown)


tzanio (approved):
The multi-touch can/should be improved, but I think it is usable in this state and worth merging. Thanks @tomstitt !

tzanio:
@kanye-quest we will like to merge this today, can you take a look?

kanye-quest (approved):
LGTM 👍

PR #161, opened 2021-03-09
Modularize shader handling

kanye-quest:
Create a `gl3::ShaderProgram` class to handle creation and binding of shaders, replacing the code in `CoreGLDevice`.

This enables other components in GLVis to use shaders in the future (e.g. order-independent transparency).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -18,7 +18,7 @@ Version 4.0.1 (development)
 
 - Various bugfixes and improvements related to the JavaScript version.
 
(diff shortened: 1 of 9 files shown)


tomstitt (approved):
This looks good to me, I added a small change to avoid a link issue with emscripten. The web build seems okay but there is something wrong in the existing master so I can't test everything.

tzanio:
I am merging this, but we need to resolve the issue from https://github.com/GLVis/glvis/issues/162

PR #163, opened 2021-03-15
ProjectVectorFEGridFunction ownership fix

### LINKED ISSUE #162: std::unique_ptr::reset causes runtime error
https://github.com/GLVis/glvis/pull/153 changed the `stream_state` to use a `std::unique_ptr`s for, among other things, the `GridFunction` `grid_f`. In the web target the `std::unique_ptr::reset` in https://github.com/GLVis/glvis/blob/master/lib/aux_js.cpp#L193 causes a runtime error that prevents the loading of example 3:

```
[Log] Switching to order 1 discontinuous vector grid function... (glvis.js, line 2293)
[Error] Unhandled Promise Rejection: RuntimeError: call_indirect to a signature that does not match (evaluating 'invoker(fn, arg0Wired, arg1Wired, arg2Wired, arg3Wired)')
	(anonymous function)
	promiseReactionJob
```

Removing the `reset` and passing the pointer to the `VisualizationSceneVector3d` "fixes" the issue in that the rendering actually happens. Here's the diff:

```
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -190,8 +190,8 @@ bool startVisualization(const std::string input, const std::string data_type,
          {
             GridFunction* proj_grid_f = ProjectVectorFEGridFunction(
                                            stream_state.grid_f.get());
-            stream_state.grid_f.reset(proj_grid_f);
-            vs = new VisualizationSceneVector3d(*stream_state.grid_f);
+            //stream_state.grid_f.reset(proj_grid_f);
+            vs = new VisualizationSceneVector3d(*proj_grid_f);
          }
          else
```

This seems like a problem with either Emscripten or how we're building the web target but I wanted to start a discussion here.

kanye-quest:
`ProjectVectorFEGridFunction` originally handled the deletion of the incoming GridFunction before #153 was merged. Afterwards, however, if the incoming GridFunction was not replaced and the original grid function is returned instead, the subsequent call to `stream_state.grid_f.reset(proj_grid_f)` deletes the original grid function.

In this PR:
- `ProjectVectorFEGridFunction` now explicitly takes ownership of the GridFunction by accepting an `std::unique_ptr<GridFunction>` by-value. It handles freeing the original grid function if a new one was generated.
- Extra declarations/definitions of `ProjectVectorFEGridFunction` were deleted.

Fixes #162.

### DIFF
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -93,8 +93,6 @@ int ReadInputStreams(StreamState& state);
 
 void CloseInputStreams(bool);
 
-GridFunction *ProjectVectorFEGridFunction(GridFunction*);
-
 // Visualize the data in the global variables mesh, sol/grid_f, etc
 // 0 - scalar data, 1 - vector data, 2 - mesh only, (-1) - unknown
 bool GLVisInitVis(int field_type)
@@ -210,9 +208,8 @@ bool GLVisInitVis(int field_type)
       {
          if (stream_state.grid_f)
          {
-            GridFunction* proj_grid_f = ProjectVectorFEGridFunction(
-                                           stream_state.grid_f.get());
-            stream_state.grid_f.reset(proj_grid_f);
+            stream_state.grid_f
+               = ProjectVectorFEGridFunction(std::move(stream_state.grid_f));
             vs = new VisualizationSceneVector3d(*stream_state.grid_f);
          }
          else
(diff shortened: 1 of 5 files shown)


tzanio (approved):
Thanks for the quick fix Max!

tomstitt (approved):
Thanks for the quick fix @kanye-quest !!

PR #168, opened 2021-04-20
Bugfixes for shader handling

kanye-quest:
Fixes a segfault identified by @dylan-copeland, caused by the calling of `glBindFragDataLocation` on an OpenGL version lacking that function.

Also:
- Corrects version checks for when to place in-shader layout specifications for outputs
- Fixes multiple render targets for WebGL 1 by adding a required `#extension GL_EXT_draw_buffers` line
- Replaces some #ifdef-ed sections with a check against a `ShaderProgram::glsl_is_es` flag

### DIFF
--- a/lib/gl/shader.cpp
+++ b/lib/gl/shader.cpp
@@ -185,8 +185,11 @@ std::string ShaderProgram::formatShader(const std::string& inShader,
 
    if (glsl_is_es)
    {
-      // Add precision specifier - required for WebGL
-      formatted = "precision mediump float;\n" + formatted;
+      if (shaderType == GL_FRAGMENT_SHADER)
+      {
+         // Add precision specifier - required for WebGL fragment shaders
+         formatted = "precision mediump float;\n" + formatted;
+      }
       if (num_outputs > 1 && glsl_version == 100)
       {
          // Enable WEBGL_draw_buffers in the shader
--- a/lib/gl/shader.cpp
+++ b/lib/gl/shader.cpp
@@ -12,13 +12,6 @@
 #include "shader.hpp"
 #include <regex>
 
-#ifdef __EMSCRIPTEN__
-// TODO: for webgl glsl the version ends in "es": #version GLSL_VER es
-const std::string GLSL_HEADER = "precision mediump float;\n";
-#else
-const std::string GLSL_HEADER = "#version GLSL_VER\n";
-#endif
-
 namespace gl3
 {
 
@@ -59,6 +52,11 @@ bool ShaderProgram::create(std::string vertexShader,
 }
 
 int ShaderProgram::glsl_version = -1;
+#ifndef __EMSCRIPTEN__
+const bool ShaderProgram::glsl_is_es = false;
+#else
+const bool ShaderProgram::glsl_is_es = true;
+#endif
 
 void ShaderProgram::GetGLSLVersion()
 {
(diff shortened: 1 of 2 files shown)


tzanio (approved):

dylan-copeland (approved):
Thanks @kanye-quest! I did not review the changes, as I have never looked at the glvis code at all, but I did test on LC and a MacBook, so my approval simply means I tested it successfully.

tzanio:
@tomstitt, do you want to take a look?

lib/gl/shader.cpp
```diff
@@ -181,16 +182,28 @@ std::string ShaderProgram::formatShader(const std::string& inShader,
...
-                                  std::to_string(glsl_version)) + formatted;
-#ifdef __EMSCRIPTEN__
-   // special prepend for WebGL 2 shaders
-   if (glsl_version == 300)
+
+   if (glsl_is_es)
    {
-      formatted = "#version 300 es\n" + formatted;
+      // Add precision specifier - required for WebGL
+      formatted = "precision mediump float;\n" + formatted;
```
tzanio, line 189: What does `precision mediump float` mean?

tzanio:
Is this worth mentioning in `CHANGELOG`?

kanye-quest:
>Is this worth mentioning in CHANGELOG?

I don't think this PR adds anything new when compared to `glvis-4.0`. This just fixes the refactored shader handling code that will be included for the next release.

lib/gl/shader.cpp
kanye-quest: They specify the precision of declared `floats` in the shader. `precision highp float` specifies a minimum 16-bit mantissa, `mediump` has 10 bits, and `lowp` has 8 bits.

The reason we need this for WebGL is because OpenGL ES doesn't specify a default precision for fragment shader variables. Also, support for `precision highp` in the fragment shader isn't guaranteed on all platforms, especially mobile devices.

Although, we should probably make this addition conditional on `if (shaderType == GL_FRAGMENT_SHADER)`, since we do want high-precision FP operations in the vertex shader.

tomstitt:
The first thing I see is that on Firefox loading the `Snake` example takes 10x longer; ~5s vs ~.5s. It isn't slower in Safari but Safari reports that it is using `GLSL 100` while Firefox reports that it is using `GLSL 300`.

On glvis.org/live Firefox and Safari report the same things (`300` and `100` respectively) so it seems like something in the WebGL 2 changes is causing a performance hit. Perhaps the default quality is higher? I can't tell but I'm happy to drop the build somewhere for others to checkout.

For `Example 3` the load time is up to ~24 seconds from ~5.

kanye-quest:
I didn't notice a slowdown with either the snake example or Example 3 on `glvis-js/examples/basic.html`, with `libglvis.js` built with optimizations; they both loaded in under a second.

For the debug version, loading an example does take on the order of seconds, but it seems most of the time is still tied up in the mfem reading and `Prepare*()` functions; I didn't see anything WebGL2-related that might be responsible for a performance slowdown:

image

@tomstitt , could you send your `libglvis.js` file and/or build options you used?

tomstitt:
> I didn't notice a slowdown with either the snake example or Example 3 on `glvis-js/examples/basic.html`, with `libglvis.js` built with optimizations; they both loaded in under a second.

Oops, it was a debug build! Sorry to waste your time on that=p

tzanio:
It is always good to double check: "measure 3 times, cut once" 😁

tzanio:
@tomstitt, do you approve or need more time to review?

tomstitt (approved):

PR #170, opened 2021-05-12
Refactoring to remove global state

kanye-quest:
This PR is intended to split off some of the work from #149 into a somewhat-smaller PR.

- Adds a new class, `GLVisWindow`, which consolidates most of the global state within `aux_vis.cpp`. `GLVisWindow` handles creation of the `SdlWindow`, `VisualizationScene*`, and `GLVisCommand` objects from a `StreamState` object and/or an array of input streams.
- Make `VisualizationScene*` event handlers member functions where possible, removing dependence on global pointers.
- Remove most references to global objects from the core glvis library.
- Move handling of multiplexing of `GLVisCommand` and other idle function runs from `SdlWindow` to `GLVisWindow`.


lib/aux_js.cpp
```diff
@@ -64,150 +63,34 @@ bool startVisualization(const std::string input, const std::string data_type,
...
-      else
-      {
-         SetVisualizationScene(vs, 3);
-      }
-   }
+   delete vs;
+   vs = nullptr;
 
-   CallKeySequence(stream_state.keys.c_str());
+   double mesh_range = -1.0;
```
tomstitt, line 86: not needed?

lib/font.hpp
```diff
@@ -62,7 +72,9 @@ class GlVisFont
    GlVisFont()
       : init(false),
         font_init(false),
-        face_has_kerning(false)
+        face_has_kerning(false),
+        is_hidpi(false),
+        ppi_w(96), ppi_h(96)
```
tomstitt, line 77: (minor) these could be default values when the members are declared:

```
... etc ...
bool face_has_kerning = false;
bool is_hidpi = false;
int ppi_w = 96, ppi_h = 96;
```

tomstitt:
I can now build and run the js version but there are a few things that aren't working yet. I saw the the `CallKeySequence` was moved from `aux_js.cpp:startVisualization` into `aux_vis.cpp:InitVisualization` (which is called in `aux_js.cpp:startVisualization`) but it doesn't seem to work, when I open ex9 I see this:

Screen Shot 2021-06-01 at 5 47 14 PM

Instead of this:

Screen Shot 2021-06-01 at 5 47 31 PM

I'll spend some more time debugging.

tomstitt:
The JS target seems to work well for me now, I slacked a copy to @tzanio and @kanye-quest so check it out if you want

tomstitt (approved):

tzanio:
All other examples from glvis.org/live work for me, but I still get an error with Navier in this branch, e.g. the following run:

```
./glvis -saved ~/glvis-web/src/data/navier.saved
```

shows

Screen Shot 2021-06-15 at 5 18 24 PM

instead of

Screen Shot 2021-06-15 at 5 18 39 PM

glvis.cpp
```diff
@@ -105,157 +105,38 @@ bool GLVisInitVis(int field_type)
...
+   try
    {
-      cerr << "Initializing the visualization failed." << endl;
-      return false;
+      mainWindow = new GLVisWindow(win_title, window_x, window_y,
+                                   window_w, window_h, legacy_gl_ctx);
    }
-
-   if (input_streams.Size() > 0)
+   catch (std::runtime_error& ex)
```
tzanio, line 113: I think this may be the first use of exceptions in MFEM or GLVis 😁

glvis.cpp
```diff
@@ -483,12 +367,9 @@ void ExecuteScriptCommand()
...
          cout << "Script: screenshot: " << flush;
 
-         if (Screenshot(word.c_str(), true))
-         {
-            cout << "Screenshot(" << word << ") failed." << endl;
-            done_one_command = 1;
-            continue;
-         }
+
+
```
tzanio, line 371: one empty line

tzanio (commented):
Preliminary review

PR #171, opened 2021-05-14
Removing POSIX dependency from GLVisCommand

kanye-quest:
Constructing a `pipe()` is needed on the X11 side in order to poll for both WM events and incoming stream commands while minimizing CPU usage. However, on Mac and Windows, we need to use their WM-native event systems to wait for events without burning CPU cycles.

This PR removes the dependency on `pipe()` and other POSIX things in `GLVisCommand`:
-  Read-after-write consistency is now enforced with an atomic `command_ready` flag.
- `SdlX11Platform` now holds an internal pipe to poll along with the window event pipe. POSIX is now only required when building with X11 support.

### DIFF
--- a/lib/sdl_x11.hpp
+++ b/lib/sdl_x11.hpp
@@ -84,8 +84,8 @@ public:
 
    ~SdlX11Platform()
    {
-       close(event_pfd[0]);
-       close(event_pfd[1]);
+      close(event_pfd[0]);
+      close(event_pfd[1]);
    }
 
    void WaitEvent()
@@ -113,19 +113,19 @@ public:
       // Read out the pending GLVisCommand-sent events, if any
       do
       {
-          char c[10];
-          n = read(event_pfd[0], c, 10);
+         char c[10];
+         n = read(event_pfd[0], c, 10);
       }
       while (n > 0);
    }
    void SendEvent()
    {
-       char c = 's';
-       if (write(event_pfd[1], &c, 1) != 1)
-       {
+      char c = 's';
+      if (write(event_pfd[1], &c, 1) != 1)
+      {
          perror("write()");
          exit(EXIT_FAILURE);
-       }
+      }
    }
 
 private:
--- a/lib/sdl_x11.hpp
+++ b/lib/sdl_x11.hpp
@@ -121,7 +121,7 @@ public:
    void SendEvent()
    {
        char c = 's';
-       if (write(pfd[1], &c, 1) != 1)
+       if (write(event_pfd[1], &c, 1) != 1)
        {
          perror("write()");
          exit(EXIT_FAILURE);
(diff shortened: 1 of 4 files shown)


tzanio (approved):
I tried it and everything looks good. Thanks @kanye-quest!

tzanio:
@tomstitt, can you take a look?

lib/sdl_x11.hpp
```diff
@@ -69,35 +71,67 @@ class SdlX11Platform final : public SdlNativePlatform
...
       while (nstr == -1 && errno == EINTR);
 
       if (nstr == -1) { perror("poll()"); }
+
+      int n = 0;
+      // Read out the pending GLVisCommand-sent events, if any
+      do
+      {
+         char c[10];
+         n = read(event_pfd[0], c, 10);
```
tomstitt, line 117: (minor) replace `10` with something like `constexpr int size = 10` or `sizeof(c)` for the 2nd `10`?

lib/sdl_x11.hpp
```diff
@@ -69,35 +71,67 @@ class SdlX11Platform final : public SdlNativePlatform
...
+      }
+      while (n > 0);
+   }
+   void SendEvent()
+   {
+      char c = 's';
+      if (write(event_pfd[1], &c, 1) != 1)
```
tomstitt, line 124: (minor) maybe we'll never use anything but a char here but replacing the `1`s with `sizeof(c)` might be better?

lib/threads.hpp
```diff
@@ -59,6 +58,8 @@ class GLVisCommand
       PALETTE_REPEAT = 20
    };
 
+   std::atomic<bool> command_ready{false};
```
tomstitt, line 61: oh neat I didn't know there were atomics in the standard library

lib/sdl_x11.hpp
```diff
@@ -69,35 +71,67 @@ class SdlX11Platform final : public SdlNativePlatform
...
+   ~SdlX11Platform()
+   {
+      close(event_pfd[0]);
+      close(event_pfd[1]);
+   }
+
    void WaitEvent()
    {
-      int nstr, nfd = 1;
+      int nstr, nfd = 2;
```
tomstitt, line 93: (minor) `constexpr`? Then `nfd` can be used instead of `2`

lib/sdl_x11.hpp
```diff
@@ -69,35 +71,67 @@ class SdlX11Platform final : public SdlNativePlatform
...
-         nfd = 2;
-      }
+
+      pfd[1].fd     = event_pfd[0];
+      pfd[1].events = POLLIN;
+      pfd[1].revents = 0;
+
       do
       {
          nstr = poll(pfd, nfd, -1);
```
tomstitt, line 106: I found my answer in the man page but if others are interested on what happens here since `pdf[1]` is setup with `O_NONBLOCK` and `poll` is called with `timeout=-1` so it is blocking:

```
The poll() function shall not be affected by the O_NONBLOCK flag.
```

tomstitt (approved):
I tested with an emscripten build and things look good to me!

PR #176, opened 2021-06-15
[DO NOT MERGE] Regression testing + screenshot fix

### LINKED ISSUE #124: Add CI testing with GitHub Actions
We should add smoke test for

- building on Linux, Mac
- code styling
- ...

kanye-quest:
Adds initial regression testing based on stream-based screenshot generation.

Also, in the process of getting this working, I discovered that some OpenGL implementations (namely, Mesa's llvmpipe software renderer) don't support reading from the front buffer. To fix this, an expose event/call is issued before taking the screenshot, so that the `Screenshot()` call grabs the back buffer data instead.

Resolves #124


kanye-quest:
@tzanio, you will probably want to squash-merge this when the time comes, otherwise the baseline images and save files from the intermediate commits might take up space in the main repo.

tzanio:
@kanye-quest, can you check Slack when you get a chance?

tzanio (commented):
This is really awesome stuff, thanks @kanye-quest!

I tried

```
python3 glvis_driver.py -e ../glvis -s ./data/streams/shaper.saved -b ./data/baselines/ -a "-lw 1 -mslw 1"
```
in the `test` directory, but I get
```
line 294, in check_shape_equality
    raise ValueError('Input images must have the same dimensions.')
```
What am I doing wrong?

A few questions and suggestions:

- The submodule is not required for the regular build, right (it is only needed in the Actions or if you want to run testing manually)?
- Consider adding a README in `test` clarifying the above, the Python requirements (e.g. install `scikit-image`), how the images are compared
- What is `PSNR`?
- Please document in CHANGELOG
- This is minor, but the `mfem` directory is called `tests` can we use that instead of `test`
- What is `xvfb`?

kanye-quest:
Regarding your issue, what are the dimensions of the `test.shaper.saved.png` image generated in the `test` directory? I think I experienced this once where the request to resize the window is limited by the actual screen dimensions. Although it should only resize to 800x600...

On the other questions:

> The submodule is not required for the regular build, right (it is only needed in the Actions or if you want to run testing manually)?

That is correct. I believe `git clone` should only pull glvis itself, and you will have to grab the `data` submodule with `git submodule update --init`.

> What is PSNR?

PSNR is the [peak signal-to-noise ratio](https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio) between two images. My only real explanation for why I went with this is "I tried the other metrics, and this one was the least worst."

It seems there are differences in rasterization between OpenGL implementations that become dramatic when you render something that is heavily affected by it, namely line meshes. So on all of the streams which enable a line mesh of some sort, we get really big differences wherever the lines are. This will be a problem with pretty much every metric that relies on pixel-to-pixel differencing.

Metrics like (structural similarity)[https://en.wikipedia.org/wiki/Structural_similarity] ostensibly help deal with cases of small differences between images. However, in some cases I was seeing poorer correlation measures between two images from the same stream, than between two images from completely different streams.

> What is `xvfb`?

`xvfb` starts an X server on headless machines, that we can then direct our applications to run on. This lets us generate screenshots on the linux gh-actions instances.

tzanio:
> Regarding your issue, what are the dimensions of the test.shaper.saved.png image generated in the test directory? I think I experienced this once where the request to resize the window is limited by the actual screen dimensions. Although it should only resize to 800x600...

It is `1600 × 1200`

tzanio:
> > The submodule is not required for the regular build, right (it is only needed in the Actions or if you want to run testing manually)?
> 
> That is correct. I believe `git clone` should only pull glvis itself, and you will have to grab the `data` submodule with `git submodule update --init`.

Ok, I confirmed that.

tzanio:
> > What is PSNR?
> 
> PSNR is the [peak signal-to-noise ratio](https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio) between two images. My only real explanation for why I went with this is "I tried the other metrics, and this one was the least worst."
> 
> It seems there are differences in rasterization between OpenGL implementations that become dramatic when you render something that is heavily affected by it, namely line meshes. So on all of the streams which enable a line mesh of some sort, we get really big differences wherever the lines are. This will be a problem with pretty much every metric that relies on pixel-to-pixel differencing.
> 
> Metrics like [structural similarity](https://en.wikipedia.org/wiki/Structural_similarity) ostensibly help deal with cases of small differences between images. However, in some cases I was seeing poorer correlation measures between two images from the same stream, than between two images from completely different streams.
> 
> >

Thanks for the explanation. Maybe add some/all of these details in the README and/or the test driver?

tzanio:
> > What is `xvfb`?
> 
> `xvfb` starts an X server on headless machines, that we can then direct our applications to run on. This lets us generate screenshots on the linux gh-actions instances.

I did not know about this and other folks may not know about it either. Can you add a brief comment in the workflow file?

tzanio:
@adrienbernede, any chance you can take a look at this? I'd like to merge it soon.

adrienbernede:
OK I'll have a look today.

tzanio (approved):
Approved modulus addressing https://github.com/GLVis/glvis/pull/176#issuecomment-872504251, https://github.com/GLVis/glvis/pull/176#issuecomment-873634696 and https://github.com/GLVis/glvis/pull/176#issuecomment-873634696.

Looks awesome, thanks for adding it @kanye-quest !

tests/glvis_driver.py
```diff
@@ -0,0 +1,144 @@
...
+    "cut_plane_transl_back": "izzzzZZZZ",
+    "orient2d_1": "R",
+    "orient2d_2": "RR",
+    "orient2d_3": "RRR",
+    "orient2d_4": "RRRR",
+    "orient2d_5": "RRRRR",
+    "orient2d_6": "RRRRRR",
+    "orient3d": "Rr",
+    "perspective": "j",
+}
```
tzanio, line 26: Can you clarify how these are used?

kanye-quest: Those were originally going to be used as input to the `-keys` command-line argument to test operations on raw meshes, but since we're only testing stream files in this PR, we could probably remove that for now.

tzanio: Or you can keep it but comment that it is not used currently

kanye-quest: Done

tzanio:
I can confirm that 8a8333c and 179f00e resolved the outstanding issues for me. Thanks @kanye-quest !

adrienbernede (commented):
I think this is an awesome contribution!
I am worried that the threshold for the image comparison (PSNR) is so low that it would not catch much, though.

I discussed this with @kanye-quest, and rasterization issue is mostly happening between OSes.
I would like to suggest 2 improvements to make these tests more robust:
- Have 2 sets of streams, one generated with with linux actions, and one generated with macos actions. This way, in CI context we compare images generated on the exact same system, which will allow to be more selective on the similarity level between images (no rasterization isssue).
- From a developer perspective. Those two sets should also reduce the occurrence on issues, but not prevent them completely. The best way to do that would then be to document how to produce a set of reference streams locally from glvis@master and use it for local testing. This could be done in a second time.

tzanio:
Thanks for the review @adrienbernede!

I am OK with having OS-specific baselines now that they are in a separate repo. @kanye-quest, what do you think?

adrienbernede:
I would also suggest that we the test suite contains at least one expected to fail test, in which we compare two similar images, but not the same, to demonstrate that we can differentiate them in the test.

tzanio:
Can you please resolve the conflicts?

kanye-quest:
Closing in favor of #189

tzanio:
Remember to delete this branch once #189 is merged

PR #179, opened 2021-07-20
Fix for building with MFEM+gslib

kanye-quest:
@brendankeith identified an issue where a forward-declared `struct array` in mfem with gslib enabled conflicted with the use of `std::array` in `palettes.cpp`. This is fixed with a header include change.

### DIFF
--- a/lib/palettes.cpp
+++ b/lib/palettes.cpp
@@ -10,7 +10,7 @@
 // CONTRIBUTING.md for details.
 
 #include "palettes.hpp"
-#include "aux_vis.hpp"
+#include "gl/renderer.hpp"
 
 #include <cmath>
 #include <cstdio>


brendankeith (approved):

psocratis (approved):

tzanio (approved):
That's probably an OK fix, but we should really look into fixing the underlying MFEM issue, cc: @kmittal2

kmittal2:
> That's probably an OK fix, but we should really look into fixing the underlying MFEM issue, cc: @kmittal2

Will be addressed in [mfem#2418](https://github.com/mfem/mfem/pull/2418).

tzanio:
Do we want this fix regardless of mfem#2418?

tzanio:
With https://github.com/mfem/mfem/pull/2418 merged, do we still need this?

kanye-quest:
There might be some value in using a header that doesn't pull in as much code, but I think we can take it or leave it at this point.

PR #174, opened 2021-06-05
Threaded server/display mode (for real this time)

kanye-quest:
# Summary

- Uses threads instead of `fork()` to open new windows for an incoming session. Each thread is associated with a single window and problem state, and handles scene operations and rendering to OpenGL.
  - As a stopgap pending #170, much of the global scene state has been made `thread_local`, and is set by a session worker thread upon a call to `GLVisInitVis()`/`GLVisStartVis()`.
  - This resolves the issue seen in #141, where Cocoa doesn't play nice with `fork()`.
  - Also potentially provides a solution for #128, since closing the server process should close all opened windows.
- Adds an `SdlMainThread` class to deal with operations that need to be handled on the main thread, such as event handling and window operations. This class is instantiated once for a program, and essentially operates as a "server" which accepts commands from session threads.
  - This is required for Windows and Mac OS X; both systems only expect a single thread to handle events. Worse, Cocoa expects this to be the main program thread only.
- Splits server code into a new function, `GLVisServer()`, which runs on its own thread.
- Enables preliminary support for building natively on Windows using CMake.

# TODO

 - [x] Get glvis-js working with these changes
 - [x] Testing for issues

PR #184, opened 2021-08-10
Create Mac application bundle

pazner:
Now that #174 was merged and GLVis behaves more like a standard Mac application (only once instance is running, single Dock icon vs. many icons, etc.), it might make sense to also package GLVis as a Mac application bundle ("GLVis.app") that can be double-clicked, added to the Dock, and so on.

This PR adds a makefile target (`make app`) that does that. It does require adding an icns file (~ 960 KB) to the repo.

image


tzanio:
This is so cool!!

kanye-quest:
Very neat @pazner! Had a few comments/suggestions:

- Can we add a similar target to the CMake build system as well?
- It might be worth considering generating both the SDL icon and the app bundle icon from a common image file
- Do we need to add code-signing? IIRC there were some requirements on newer versions of macOS that required notarization for app bundles (although maybe this only is needed for distribution)

pazner:
> * Can we add a similar target to the CMake build system as well?

Yes, I just added a CMake target. `make app` should work with CMake now as well.

> * It might be worth considering generating both the SDL icon and the app bundle icon from a common image file

I agree, I can generate the icns file from the SDL icon with ImageMagick and iconutil. If we had a PNG format icon instead of raw RGBA data, we could generate the icns using only iconutil (which comes standard with Mac OS).

```bash
mkdir -p GLVis.iconset
convert -size 1024x1024 -depth 8 logo.rgba GLVis.iconset/icon_512x512@2x.png
iconutil --convert icns GLVis.iconset
rm -r GLVis.iconset
```

> * Do we need to add code-signing? IIRC there were some requirements on newer versions of macOS that required notarization for app bundles (although maybe this only is needed for distribution)

I don't think you need code-signing to run an app that you compile yourself, only to distribute it. It would be good to test this though.

pazner:
@kanye-quest, see the branch [mac-app-icon](https://github.com/glvis/glvis/tree/mac-app-icon) for an example of creating the icns file directly from the RGBA data. Instead of ImageMagick, I used the Mac's AppKit APIs to convert to PNG. Since if we are building a Mac app bundle, it's probably OK to depend on having Mac SDKs. That being said, it might be more hassle than it's worth compared with just including the ~ 1 MB icon file.

tzanio:
If you don't want to see the `.app` extension, deselect "Show all filename extensions" in Finder > Preferences > Advanced

Is it possible to add something to the dialog for the "About GLVis" menu item? Currently it only shows the GLVis logo, it will be nice to at least link to the website.

pazner:
> Is it possible to add something to the dialog for the "About GLVis" menu item? Currently it only shows the GLVis logo, it will be nice to at least link to the website.

Yep, the contents of the file `Credits.rtf` will be displayed in the "About" window:

Screen Shot 2021-08-10 at 10 06 48 PM

tzanio:
> Yep, the contents of the file Credits.rtf will be displayed in the "About" window:

Thanks @pazner! I adjusted the text a bit -- feel free to edit further if you'd like.

Screen Shot 2021-08-11 at 8 50 24 AM

tzanio (approved):
This is awesome, thanks for thinking about it and making it happen @pazner 🎉 

- [x] Please mention in `CHANGELOG`

tzanio:
@kanye-quest and @tomstitt -- do you need more time to review this?

(totally fine if you do, just checking)

tomstitt (approved):
I just wanted to try it out! Very cool, nice work @pazner !

kanye-quest (approved):
Unfortunately, I don't have a Mac to test this with for the time being, so I'm deferring to @tomstitt and @tzanio. But the changes look good to me!

PR #185, opened 2021-08-11
Fix for live site hang

kanye-quest:
In the live website, an [event handler](https://github.com/GLVis/web/blob/69df65dc585f60adafcfe131214904572b881fa1/src/live/index.html#L740) which calls `SdlWindow::SetWindowSize()` is attached to the window resize event. However, this may be called before the SdlWindow is actually initialized; the end result is that the live site hangs on a `condition_variable` that is never woken. To fix this:
- We call `SetSingleThread()` earlier in the `SdlMainThread` initialization, so we can avoid calling multithreaded code before the first window is initialized.
- Window-specific functions on `SdlMainThread` first check for the validity of the incoming handle, in order to avoid queuing up events for a null window.

### DIFF
--- a/lib/sdl.cpp
+++ b/lib/sdl.cpp
@@ -92,7 +92,6 @@ bool SdlWindow::createWindow(const char* title, int x, int y, int w, int h,
 {
 #ifdef __EMSCRIPTEN__
    is_multithreaded = false;
-   GetMainThread().SetSingleThread();
 #endif
    // create a new SDL window
    handle = GetMainThread().GetHandle(this, title, x, y, w, h, legacyGlOnly);
--- a/lib/sdl_main.cpp
+++ b/lib/sdl_main.cpp
@@ -73,6 +73,9 @@ SdlMainThread::SdlMainThread()
                << "." << (int)sdl_ver.patch << std::endl);
 
    sdl_init = true;
+#ifdef __EMSCRIPTEN__
+   SetSingleThread();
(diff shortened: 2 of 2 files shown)


tzanio:
@tomstitt or @v-dobrev, were you able to give this a try?

tzanio:
@kanye-quest, this does not fix the font issue from #182, correct?

tomstitt:
> @kanye-quest, this does not fix the font issue from #182, correct?

I believe that is fixed in a PR beyond the commit of glvis I ultimately pushed into glvis-js. I will test this PR and verify!

v-dobrev (approved):
I don't have the setup to test the js build, so I only tested a build on macOS.

tomstitt:
I tested this and `glvis/web` works for me with `mkdocs serve`! awesome work @kanye-quest !!

I see the small font issue though so I'm not sure what I was seeing before when I thought it was fixed in `master`.

tomstitt (approved):

tzanio (approved):
Thank you @kanye-quest 💯

tzanio:
Sorry to bug you @tomstitt, but can you regenerate the library in the glvis-js repo?

tomstitt:
> Sorry to bug you @tomstitt, but can you regenerate the library in the glvis-js repo?

done!

tzanio:
thanks!

PR #188, opened 2021-08-14
Update numbering when a new mesh and solution is read

rw-anderson:
Fix to update numberings when a new mesh and solution is read in. This is important, for example, when the new mesh is adaptively refined.

### DIFF
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -592,6 +592,7 @@ void VisualizationSceneSolution::NewMeshAndSolution(
    PrepareLevelCurves();
    PrepareBoundary();
    PrepareCP();
+   PrepareNumbering();
    PrepareOrderingCurve();
 }
 
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -1564,16 +1564,6 @@ double VisualizationSceneSolution::GetElementLengthScale(int k)
 
 void VisualizationSceneSolution::PrepareElementNumbering()
 {
-   int ne = mesh -> GetNE();
-
-   if (ne > MAX_RENDER_NUMBERING)
-   {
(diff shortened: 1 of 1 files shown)


tzanio:
Is this `ready-for-review` @rw-anderson ?

tzanio (approved):
Thanks Bob!

rw-anderson:
An unfortunate side effect of this change is that the warning about rendering for large meshes is spit out every time a new mesh and solution is read in.
I suppose with the work that's been done lately, I should test to make sure that's still a performance issue.
And secondly, if it still is, to only emit the warning when the numbering is requested with a keypress, not every time it goes through the pre-computation phase.

kanye-quest:
With the shader-based renderer (macOS, Linux, and WebGL clients), there shouldn't be a significant performance impact; we upload the text as a vertex buffer which is then transformed to the correct position inside the shader. There may be a performance impact with the legacy OpenGL renderer (remote X11 or with `-oldgl` switch), since we have to issue draw calls for each triangle in the text.

kanye-quest (approved):

tzanio:
@rw-anderson can you test and let us know how best to resolve https://github.com/GLVis/glvis/pull/188#issuecomment-899101105?

I'll hold on merging this until then.

tzanio:
@rw-anderson any progress?

(We want to have most of the glvis-4.1 PRs merged by the end of this week)

rw-anderson:
Sorry I haven't looked at this again yet. I should be able to take a look on Friday.

tzanio:
Friday is OK, thanks!

rw-anderson:
I tested it and performance seems fine with the latest code, so I'm just removing the mesh size limitation.

PR #190, opened 2021-08-18
Minor cleanup of thread signaling

kanye-quest:
- For `SdlWindow::signalLoop()` and `SdlWindow::signalKeyDown()`: instead of waking the main thread, which then wakes up the worker threads, we just signal the target worker thread directly.
  This avoids waking up all the worker threads to handle just a single thread's pending stream-based event. (i.e. the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem))
- Removes the no-longer-needed custom SDL event mechanism.
- Removes an unused parameter from `SdlMainThread::CreateWindowCmd`.

### DIFF
--- a/lib/threads.hpp
+++ b/lib/threads.hpp
@@ -26,6 +26,7 @@ private:
    VisualizationSceneScalarData **vs;
    StreamState&         curr_state;
    bool                 *keep_attr;
+   SdlWindow            *thread_wnd;
 
    std::mutex glvis_mutex;
    std::condition_variable glvis_cond;
--- a/lib/sdl.cpp
+++ b/lib/sdl.cpp
@@ -626,7 +626,7 @@ void SdlWindow::signalKeyDown(SDL_Keycode k, SDL_Keymod m)
       event.key.keysym.sym = k;
       event.key.keysym.mod = m;
    }
-   SDL_PushEvent(&event);
+   queueEvents({ event });
 }
 
 void SdlWindow::swapBuffer()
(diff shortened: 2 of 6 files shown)


v-dobrev (approved):
I tried it on Mac and did not notice any issues.

tzanio (approved):

tzanio:
@tomstitt, are you interested in taking a look?

lib/threads.hpp
```diff
@@ -26,6 +26,7 @@ class GLVisCommand
    VisualizationSceneScalarData **vs;
    StreamState&         curr_state;
    bool                 *keep_attr;
+   SdlWindow            *thread_wnd;
```
tomstitt, line 29: this probably doesn't matter but if `GLVisCommand` is default-constructed and `signal` is called `thread_wnd` might not be `nullptr` which would fault, consider `nullptr` initializing `thread_wnd` or hide/delete the default ctor

tomstitt:
I tried the web build and it works with glvis-web and `mkdocs serve`

tomstitt (approved):

lib/threads.hpp
kanye-quest: I think the default constructor for `GLVisCommand` is implicitly hidden by our defined constructor.

tomstitt: You're right! I should have looked it up, sorry about that

ref: https://en.cppreference.com/w/cpp/language/default_constructor

PR #192, opened 2021-08-20
Rename Github Actions, needs update of mfem-actions version

adrienbernede:
Looking good. Ready for review and merge.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -52,7 +52,7 @@ jobs:
             target: dbg
             mpi: par
             build-system: make
-    name: ${{ matrix.os }}-${{ matrix.target }}-${{ matrix.mpi }}-${{ matrix.build-system }}
+    name: ${{ matrix.os }}-${{ matrix.build-system }}-${{ matrix.target }}-${{ matrix.mpi }}
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -32,25 +32,25 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-10.15]
-        target: [debug, optim]
-        mpi: [sequential]
+        target: [dbg, opt]
+        mpi: [seq]
         build-system: [cmake]
         include:
           - os: ubuntu-20.04
-            target: optim
-            mpi: sequential
+            target: opt
+            mpi: seq
             build-system: make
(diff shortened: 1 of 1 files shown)


tzanio:
Seems to be enough?

.github/workflows/builds.yml
```diff
@@ -32,25 +32,25 @@ jobs:
...
+            target: opt
+            mpi: seq
             build-system: make
           - os: macos-10.15
-            target: debug
-            mpi: parallel
+            target: dbg
+            mpi: par
             build-system: make
     name: ${{ matrix.os }}-${{ matrix.target }}-${{ matrix.mpi }}-${{ matrix.build-system }}
```
tzanio, line 55: For consistency with MFEM:

```suggestion
    name: ${{ matrix.os }}-${{ matrix.build-system }}-${{ matrix.target }}-${{ matrix.mpi }}
```

.github/workflows/builds.yml
```diff
@@ -32,25 +32,25 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, macos-10.15]
-        target: [debug, optim]
-        mpi: [sequential]
+        target: [dbg, opt]
+        mpi: [seq]
```
tzanio, line 36: I'm confused by this shouldn't it be either empty or `mpi: [seq, par]`?

adrienbernede: This can be confusing, I agree.

Here is how that works:

1 - Matrix evaluation. Any field in the matrix becomes a dimension, having a dimension of 1 means that the field will be constant. Github will generate 4 jobs from this matrix. That part is easy.

2 - Include evaluation. Any include that specifies all the fields in the matrix is a new job. This can be used to add a specific scenario. Here, we use include to add a value to `build-system`: make. And we fill the other fields as we want.

3 - Include evaluation (bis). Any include that specifies only a subset of all the parameters is meant to add a parameter to the jobs matching the specified parameters.

Our definition is excluding: cmake + parallel, make + opt + par, make + dbg + seq. 
Why? I can’t really remember. I think we didn’t care much about parallel builds, but thought we would use them to do test the debug target as well.

The situation is borderline: we could as well have a flat description of the 8 jobs.

tzanio: Aha, I see. Thanks for the explanation.

tzanio (approved):

kanye-quest (approved):

PR #230, opened 2022-05-14
Support for faces with holes in 3D

tzanio:
Added the option to cut a portion of the interiors of 3D faces to expose more of the mesh. Useful as an alternative to transparency. 

See keys <kbd>Ctrl</kbd> + <kbd>F3</kbd> / <kbd>F4</kbd>.

**Example:**

Screen Shot 2022-05-13 at 5 46 51 PM

**Inspired by:**

<table>
  <tr>
    <td> </td>
    <td></td>
   </tr> 
</table>

### DIFF
--- a/lib/openglvis.cpp
+++ b/lib/openglvis.cpp
@@ -126,6 +126,9 @@ VisualizationScene::VisualizationScene()
    ViewCenterX = 0.0;
    ViewCenterY = 0.0;
 
+   cut_lambda = 0.0;
+   cut_updated = false;
+
    background = BG_WHITE;
    GetAppWindow()->getRenderer().setClearColor(1.f, 1.f, 1.f, 1.f);
    _use_cust_l0_pos = false;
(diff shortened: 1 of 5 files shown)


v-dobrev (approved):

tzanio:
@mlstowell, @kanye-quest, @dylan-copeland, and @psocratis -- any chance I can get another approval on this PR?

lib/openglvis.cpp
```diff
@@ -137,6 +140,113 @@ VisualizationScene::VisualizationScene()
...
+      c[0] += p[j][0]; c[1] += p[j][1]; c[2] += p[j][2];
+   }
+   c[0] /= 3.0; c[1] /= 3.0; c[2] /= 3.0;
+
+   double l = cut_lambda;
+   double q[3][3];
+
+   for (int j = 0; j < 3; j++)
+   {
+      q[j][0] = l*p[j][0] + (1-l)*c[0];
```
mlstowell, line 163: I would prefer to see `1.+l` or `1.0+l` and avoid the implicit type conversion.

tzanio:
> I like this feature but, unless I missed something, it looks like these cut triangles and quads are being displayed even if `cut_lambda=0`. Is that correct? If so, wouldn't it be more efficient to display a single triangle or quad?

They shouldn't be, see e.g. this check: https://github.com/GLVis/glvis/pull/230/files#diff-8fd2270828db1d1202e66618a756d214a775cd82f71d29d9a92d475485a5b32dR747-R751

Am I missing something 🤔

lib/openglvis.cpp
tzanio: ok

tzanio:
> > I like this feature but, unless I missed something, it looks like these cut triangles and quads are being displayed even if `cut_lambda=0`. Is that correct? If so, wouldn't it be more efficient to display a single triangle or quad?
> 
> They shouldn't be, see e.g. this check: https://github.com/GLVis/glvis/pull/230/files#diff-8fd2270828db1d1202e66618a756d214a775cd82f71d29d9a92d475485a5b32dR747-R751
> 
> Am I missing something 🤔

Oh I see, you mean in `PrepareFlat`... I'll fix that

tzanio:
@mlstowell, see https://github.com/GLVis/glvis/pull/230/commits/9e3ceb8c4b15434485d90bf1893b5ddfab55df49

lib/vssolution3d.cpp
```diff
@@ -2695,11 +2894,11 @@ void VisualizationSceneSolution3d::PrepareCuttingPlane2()
 
             if (nodes.Size() == 3)
             {
-               DrawTriangle(cplane_buf, p, c, minv, maxv);
+               DrawCutTriangle(cplane_buf, p, c, minv, maxv);
```
mlstowell, line 2897: Should these lines also check for `cut_lambda=0`?

lib/openglvis.cpp
mlstowell: Thanks

lib/vssolution3d.cpp
tzanio: Thanks for catching that, I removed it.

mlstowell (approved):

PR #235, opened 2022-08-03
Parallel streams for web version

tomstitt:
- Adds and renames the `display` and `update` methods
- Moves `ReadInputStreams` from `glvis.cpp` to `lib/stream_state.*`
    - `keep_attr` is now a member of `StreamState`

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -35,6 +35,8 @@ std::vector<unsigned char> * screen_state = nullptr;
 
 StreamState stream_state;
 
+int last_stream_nproc = 1;
+
 namespace em = emscripten;
 
 namespace js
@@ -263,6 +265,8 @@ int processParallelStreams(StreamState & state, const StringArray & streams,
       istreams[i].release();
    }
 
+   last_stream_nproc = streams.size();
+
    return field_type;
 }
 
@@ -331,6 +335,12 @@ int updateStream(std::string stream)
 
 int updateParallelStreams(const StringArray & streams)
 {
+   if (streams.size() != last_stream_nproc)
+   {
+      std::cout << "current stream-list size does not match the previous: " <<
+                streams.size() << " != " << last_stream_nproc << std::endl;
+      return 1;
+   }
    StreamState new_state;
    processParallelStreams(new_state, streams);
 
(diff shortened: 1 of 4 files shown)


tomstitt:
The mac jobs are failing due to `GitHub Actions has encountered an internal error when running your job.`, any ideas?

kanye-quest:
Looks like they deprecated the macos-10.15 container on gh actions? https://github.com/actions/virtual-environments/issues/5583

tzanio:
I switched in https://github.com/mfem/mfem/pull/3162 to `macos-latest`.

Should we do the same here?

lib/aux_js.cpp
```diff
@@ -222,23 +223,76 @@ bool startVisualization(const std::string input, const std::string data_type,
...
-      return 1;
+      commands->seekg(istreams[0]->tellg());
    }
 
-   StreamState new_state;
-   new_state.ReadStream(ss, data_type);
+   // HACK: don't let unique_ptr free the data
+   for (int i = 0; i < streams.size(); ++i)
+   {
+      istreams[i].release();
```
kanye-quest, line 265: Are these streams being used elsewhere outside this function?

kanye-quest (approved):

lib/aux_js.cpp
tomstitt: No they are not.

Ideally we can come up with a way to get a `std::vector<std::unique_ptr<std::istream>>` view of the `std::vector<std::string>` argument but I'm not sure how to do that. Any ideas? The other option is to change the signature of `ReadStream` but that would mean changing `glvis.cpp`

kanye-quest: I guess I was asking more about the use of `unique_ptr::release()`; I think the use of `std::stringstream` is fine given that we also have to handle the case of live sockets via mfem's `socketstream` when running as a desktop application. It also looks like the `mfem::Mesh` and `GridFunction` ctors expect an `std::istream`-like argument.

tomstitt: Ah! Good question. The `unique_ptr`s are referencing the stack allocated `streamstream`s and I don't want `unique_ptr` to try to destruct them since that would be a double-free. I'm not sure if we can improve, thoughts ?

Issue #236, opened 2022-08-15
Significant increase in memory usage


tzanio:
Something happened between `glvis-4.0` and `glvis-4.1 `that more than doubled the memory usage

If I do `glvis -m mesh.000000 -g examples/sol.000000` (these are the saved files from `mpirun -np 1 ex1p -m ../data/square-disc-p2.mesh`), then I get the following memory usage:

- glvis-3.4 🠚 0.21 GB
- glvis-4.0 🠚 0.87 GB
- glvis-4.1 🠚 2.24 GB
- glvis-4.2 🠚 2.79 GB
- glvis-now 🠚 2.80 GB

Measured on my Mac with: `ps x -o rss,vsz,command | grep glvis | grep '\/' | awk '{ print $1/1024/1024 }'`

Maybe some of this is necessary, but it will be good to check if we can reduce it, since it impacts load time and the web version (see the recent MFEM AWS [tutorial](https://mfem.org/tutorial/))

Data files:
- [mesh.000000.gz](https://github.com/GLVis/glvis/files/9339792/mesh.000000.gz)
- [sol.000000.gz](https://github.com/GLVis/glvis/files/9339794/sol.000000.gz)

kanye-quest:
#237 should take care of some of the low-hanging fruit.

Beyond that, I did some profiling of the mesh in question on a Linux system with massif. First, it seems that on Linux roughly 413MB is in use, instead of almost a full GB, so maybe some of the memory usage on Macs is related to the switch to Cocoa. 

A general summary of allocations:
- Element/vertex/edge markers are a considerable chunk of that; the constituent lines make up roughly 80MB
- Text entry buffers for those markers are 40MB - the strings themselves are 32MB
- Ordering curve cones vertex data comes in at 80MB; the indices come in at 24MB
- Ordering curve cone lines come in at 16MB
- The mesh itself uses 32MB

Note that the markers are not generated for 3D meshes; I profiled a 3D stream file and the biggest memory hog was the ordering curve.

## Things to try next

Generating markers could be deferred until the user requests that they be drawn.

At least some of this memory usage is the result of `std::vector`'s resize behavior; if we can precompute and reserve the exact amount of buffer space we need, we can reduce some of these numbers.

[...]

PR #237, opened 2022-08-19
Memory usage improvements

kanye-quest:
# Summary

- Defer buffering objects to the GPU until they're actually used in a pending frame
- Use indexed buffers to hold ordering curve
- Don't draw cones when cone_scale is 0 (`PrepareOrderingCurves()` uses `Arrow3` to draw the no-arrow curve)

With these changes, I get around 0.97GB of memory usage for the command in #236, which puts us almost at where we were at version glvis-4.0.

## Things to try:

- [ ] Remove second texture coordinate - this should drop us down from 32 to 16 bytes/vertex when storing `VertexTex` due to alignment

### DIFF
--- a/lib/vsdata.cpp
+++ b/lib/vsdata.cpp
@@ -12,6 +12,8 @@
 #include <cstdlib>
 #include <cmath>
 
+#include <unordered_set>
+
 #include <iomanip>
 #include <sstream>
 #include <limits>
@@ -986,8 +988,6 @@ gl3::SceneInfo VisualizationSceneScalarData::GetSceneObjs()
    int w, h;
    wnd->getWindowSize(w, h);
    gl3::SceneInfo scene {};
-   scene.needs_buffering = std::move(updated_bufs);
-   updated_bufs.clear();
 
    gl3::RenderParams params {};
    params.model_view.identity();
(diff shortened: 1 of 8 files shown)


tzanio:
@kanye-quest, the current failure is not due to that, but you probably want to merge `master` for its macos CI fixes

tzanio:
Should we mention it in `CHANGELOG`?

tzanio (approved):

tzanio:
@v-dobrev and @tomstitt -- can you please review when you get a chance?

lib/openglvis.cpp
```diff
@@ -261,14 +261,13 @@ ::DrawTriangle(gl3::GlDrawable& buff,
       return;
    }
 
-   std::array<float, 2> texcoord[3];
+   float texcoord[3];
    std::array<float, 3> fpts[3];
    std::array<float, 3> fnorm = {(float) nor[0], (float) nor[1], (float) nor[2]};
 
    for (int i = 0; i < 3; i++)
```
tomstitt, line 268: (minor) I don't think this will every change but consider using some constexpr instead of `3`

lib/vsdata.cpp
```diff
@@ -73,14 +75,20 @@ void VisualizationSceneScalarData::DoAutoscaleValue(bool prepare)
...
+static std::array<float, 3> ToVec3(const T* vec)
+{
+   return { (float) vec[0], (float) vec[1], (float) vec[2] };
+}
+
+void VisualizationSceneScalarData::Cone(gl3::GlDrawable& buf,
+                                        glm::mat4 xfrm,
+                                        double cval)
 {
    const int n = 8;
```
tomstitt, line 88: (minor/question) do you know if the compiler will do the same stuff with `const int n = 8` as `constexpr int n = 8`? should it change? (along with the other `const <numeric type>`)

kanye-quest: If I had to guess, they'd probably be equivalently optimizable in this case.

tomstitt (approved):
Awesome work @kanye-quest !!

lib/vsdata.cpp
tomstitt: Yes, I checked, at least with intel the output is the same in compiler explorer

tzanio (approved):

PR #238, opened 2022-08-24
Update os versions to latest

tomstitt:


adrienbernede (approved):
For the records, I personally prefer fixed versions because the change of version is then controlled and happens on one branch, which can also contain the appropriate changes/fixes.
When using "latest", we expose the code to the risk of a CI failure applying to all the branches at once.
I am guessing this was requested, but I wanted to give my 2 cents.

tzanio:
> For the records, I personally prefer fixed versions because the change of version is then controlled and happens on one branch, which can also contain the appropriate changes/fixes.
When using "latest", we expose the code to the risk of a CI failure applying to all the branches at once.
I am guessing this was requested, but I wanted to give my 2 cents.

Good points @adrienbernede.

We already made this change in MFEM (with the idea of reducing maintenance and forcing us to keep up with OS updates) so the change here is for uniformity.

Lets see how it goes, I have the feeling we are not done with tweaking the CI system :)

tzanio (approved):

PR #242, opened 2022-09-14
Fix build on big-endian

pkubaj:


tzanio (approved):
thanks for the fix @pkubaj !

bslazarov (approved):

PR #250, opened 2023-01-04
Fix GetFaceGeometry deprecation warnings

pazner:
- Use `GetFaceGeometry` instead of `GetFaceBaseGeometry`
- Use `snprintf` instead of `sprintf`


tzanio (approved):
Thanks @pazner

PR #262, opened 2023-08-17
Added command to plot levellines from stream and script.

J-Pi:
Added the command `levellines` to plot levellines from stream command and script. However, the key m needs to be pressed twice to activate the rendering of the levellines. When `autoscale` is `mesh` the levellines will be ploted for further timesteps, otherwise it has to be used the command again.

This is needed for the new Example 37 of mfem (see PR [#3828](https://github.com/mfem/mfem/pull/3828)).

### DIFF
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -634,6 +634,17 @@ void ExecuteScriptCommand()
          }
          cout << endl;
       }
+      else if (word == "levellines")
+      {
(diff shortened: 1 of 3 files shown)


vladotomov (approved):

tzanio (approved):
Thanks @J-Pi !

PR #279, opened 2024-05-23
A compilation parameter for font size

najlkin:
This PR adds a compilation parameter for the default font size (afterwards can be adjusted at runtime), which might be useful to adjust for different machines.


tzanio (approved):

termi-official (approved):
LGTM

PR #280, opened 2024-05-23
Support for integral finite elements

najlkin:
This PR adds support for integral finite elements, which were loaded, but crashed GLVis 💣 .

Status:

- [x] Fixed the crash for scalar grid functions (by turning off normals)
- [x] Compute normals for smooth surface visualization (approximated by value-based FEs)
- [ ] Adjust the refinement to respect mesh curvature -> #285 
- [ ] ~Vector grid functions in 2D~ (No smooth visualization of vector fields in 2D 😞 )
- [ ] ~Vector grid functions in 3D~ (No smooth visualization with value-based displacement 😞 )

### DIFF
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -749,7 +749,7 @@ int VisualizationSceneSolution::GetRefinedValuesAndNormals(
             {
                const IntegrationPoint &ip = nodes.IntPoint(n);
                Trans->SetIntPoint(&ip);
-               lval(n) /= Trans->Weight();
+               lval(n) /= Trans->Weight();//value = dof / |J|
             }
 
             // Gradient calculation
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -725,7 +725,7 @@ int VisualizationSceneSolution::GetRefinedValuesAndNormals(
    if (drawelems < 2)
    {
       // In 1D we do not have well-defined normals.
-      if (dim > 1)
+      if (dim > 1 && rsol->FESpace()->GetFE(i)->GetMapType() == FiniteElement::VALUE)
       {
          rsol->GetGradients(i, ir, tr);
          normals.SetSize(3, tr.Width());
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -727,11 +727,12 @@ int VisualizationSceneSolution::GetRefinedValuesAndNormals(
       // In 1D we do not have well-defined normals.
       if (dim > 1)
       {
-         if (rsol->FESpace()->GetFE(i)->GetMapType() == FiniteElement::VALUE)
+         const int map_type = rsol->FESpace()->GetFE(i)->GetMapType();
+         if (map_type == FiniteElement::MapType::VALUE)
          {
             rsol->GetGradients(i, ir, tr);
          }
-         else
+         else (map_type == FiniteElement::MapType::INTEGRAL)
          {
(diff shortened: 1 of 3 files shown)


tzanio:
Mention in `CHANGELOG`?

lib/vssolution.cpp
```diff
@@ -727,7 +727,44 @@ int VisualizationSceneSolution::GetRefinedValuesAndNormals(
       // In 1D we do not have well-defined normals.
       if (dim > 1)
       {
-         rsol->GetGradients(i, ir, tr);
+         if (rsol->FESpace()->GetFE(i)->GetMapType() == FiniteElement::VALUE)
+         {
+            rsol->GetGradients(i, ir, tr);
+         }
+         else
```
termi-official, line 734: I think we should be explicit here (i.e. `FiniteElement::INTEGRAL`) with the mapping type selection and reserve the else branch for error handling (e.g. UNKNOWN_MAP_TYPE ).

lib/vssolution3d.cpp
```diff
@@ -3928,7 +3924,47 @@ void VisualizationSceneSolution3d::PrepareLevelSurf()
          RefG = GLVisGeometryRefiner.Refine(geom, TimesToRefine);
          GridF->GetValues(ie, RefG->RefPts, vals, pointmat);
 #ifdef GLVIS_SMOOTH_LEVELSURF_NORMALS
-         GridF->GetGradients(ie, RefG->RefPts, grad);
+         if (GridF->FESpace()->GetFE(ie)->GetMapType() == FiniteElement::VALUE)
+         {
+            GridF->GetGradients(ie, RefG->RefPts, grad);
+            gp = &grad;
+         }
+         else
```
termi-official, line 3932: See above.

termi-official (commented):
Two questions left from my side (beside the comments).

1.  H(div) and H(curl) elements can be interpreted as integral elements. I assume they still crash on this branch. Can we add meaningful error messages for these?
2. Can add some comment on the local projection (in the code)? For me it is not really obvious why this is the correct way to handle the visualization.

Furthermore, is there some way to add a simple regression test on this?

lib/vssolution.cpp
najlkin: Done 👍

najlkin:
> Two questions left from my side (beside the comments).
>  1. H(div) and H(curl) elements can be interpreted as integral elements. I assume they still crash on this branch. Can we add meaningful error messages for these?

H(div) and H(curl) are vector finite elements, so they are visualized as vector fields by a completely different object. Unfortunately, GLVis does not have smooth (i.e. interpolated normals) visualization of any vector fields, as I mentioned in #284 .
 
>  2. Can add some comment on the local projection (in the code)? For me it is not really obvious why this is the correct way to handle the visualization.

I added a little comment, but it is really simple. The point value of integral elements is the DOF value divided by the Jacobian, see `MapType` documentation.

> Furthermore, is there some way to add a simple regression test on this?

Regression testing is poor in GLVis, it only checks against saved streams from runs of example codes. It does not cover a great fraction of the options. I have proposed to improve it, but it does not have a priority. Anyway, it is a different repo, so it even cannot be added here, and also we do not have exmaples with integral elements, I think.

termi-official (approved):
Thanks for elaborating @najlkin and also thanks for working on this!

tzanio (approved):
Thank you for adding this, @najlkin!

Issue #272, opened 2024-02-08
Updating GLVis mesh vector or scalar solution without drawing a new mesh


altineller:
Hello,

I am about to make a plugin for glVis that listens on a different socket, and updates the mesh node positions.

My purpose is to get an animation of displacements.

I have read the code for the b button press, which is at vsvector3d.cpp and

`
void VisualizationSceneVector3d::NPressed()
{
   if (drawdisp)
   {
      PrepareDisplacedMesh();
   }
   else
   {
      Prepare();
      PrepareLines();
   }

   SendExposeEvent();
}
`

I am intending to make a connector, that triggers a method like NPressed above, so I can edit displace mesh, or update solution.

But before I do this, has there been any work done in similar direction?

Any ideas / reccomendations / help greatly appreciated.

Best Regards,
C.

tzanio:
How is this different than what happens e.g. in ex10 by pushing updated fields to the same socket?

altineller:
Hello,

In ex10.cpp, line 381:  void visualize(ostream &os, Mesh *mesh, GridFunction *deformed_nodes, GridFunction *field, const char *field_name, bool init_vis) ..

    GridFunction *nodes = deformed_nodes;
    int owns_nodes = 0;
    mesh->SwapNodes(nodes, owns_nodes);
    os << "solution\n" << *mesh << *field;
    mesh->SwapNodes(nodes, owns_nodes);

Where the nodes of the mesh are temporarily swapped, and sent to glVis trough socket. but mesh is being resent again.

I am interesting modifying mesh inside glVis programmatically, and also pass binary data, instead of ascii, to get at least 10hz update rate, for a mesh with around 65K nodes. So I would like to pass as little as possible data.

Is there any reason for the underlying visualized mesh in glVis memory to be unmodifiable? 

Best Regards,
Can

tzanio:
Interesting 👍 

Currently when the solution is updated the mesh is also sent as if it is done for the first time. See the logic in https://github.com/GLVis/glvis/blob/master/lib/threads.cpp#L760

This and https://github.com/GLVis/glvis/blob/master/lib/threads.cpp#L425 could probably be changed, e.g. by adding a new keyword, to just send the grid function and reuse the current mesh.

Tzanio

joetristano:
Would love to have this feature!!!! Was just wondering about the same thing.  Redundant data tranfer can be easily eliminated...

Issue #268, opened 2024-01-27
Additional color palettes?


tzanio:
Consider adding support for additional color palettes such as the colour-vision deficiency friendly and perceptually-uniform colour maps at https://zenodo.org/records/8409685

![+ScientificColourMaps8-FabioCrameri](https://github.com/GLVis/glvis/assets/1278247/115bd0cf-c853-48a1-82dc-e723dca8d9d2)

See also https://github.com/GLVis/glvis/tree/colormap_dev

justinlaughlin:
I can work on this.

justinlaughlin:
I think I've tried every one of the colormaps from Fabio Crameri in grad school :'). A couple quick thoughts... colormap designers often emphasize linear luminosity (one of the main criticisms of jet is that its luminosity is nonlinear and non-monotonic) which is a feature that makes perfect sense for 2d data. However in 3d it can start to lose significance because there is almost always some kind of lighting applied to the scene. It is still a good thing to have, but IMO span of luminosity+color is more important in a scene with lighting. I like `batlow` and `lipari` from this set, as they are perceptually uniform, span many colors, and look good. I'm not a huge fan of colormaps like `buda` and `nuuk` as their range of color/luminosity is small, so scene lighting will more dramatically influence perception.

Lmk if you have any thoughts on which colormaps we should add.

tzanio:
I'm also wondering if we need more colormaps, but if we decide to add them, I do like `batlow` and `lipari` myself.

We already have some  perceptually uniform maps in https://github.com/GLVis/glvis/blob/master/lib/palettes.cpp from https://github.com/1313e/CMasher and https://bids.github.io/colormap/.

justinlaughlin:
If there is too many options we might need a different interface to choose a colormap besides "cycling". Maybe the most common colormaps can be cycled through, and for those who want more options, a separate hotkey could let you input a name to a specific colormap. That way we could add in as many colormaps as we want without cluttering the colormap cycle.

tzanio:
You can always "jump" to a particular color palette with <kbd>F6</kbd>, but I agree that cycling through a lot of palettes has its downsides.

Another option could be to enable extra palettes with a build options.

tzanio:
Addressed in https://github.com/GLVis/glvis/pull/289

Issue #260, opened 2023-06-30
Change colorbar number format


ddkn:
Hello,

Looking at the documentation I wasn't sure how to change the string format on the colorbar, or is this even possible? I would like it to be more consistent. Depending on the amount of significant figures you may get a *float* or *scientific notation* in the same colorbar.

![image](https://github.com/GLVis/glvis/assets/13819239/60af57ff-b97b-4bf0-a3df-acff9fbc6270)

It would be nice to have more control, such as show sign (+ and -) and control the significant figures after the decimal, _e.g._ +3.145e+00 and -3.145e+00. Again, I apologize if I have missed this option

Or is there a better way to export this for plotting?

Thanks in advance!

_PS_ I believe `glvis` was installed through `pip install glvis==0.3.3`, which I think pulled in the latest binary 4.2

tzanio:
There is unfortunately no way currently to control this at user level.

ddkn:
Do you think it would be a desired feature? I may look into it trying to add it once I defend.

tzanio:
> Do you think it would be a desired feature? I may look into it trying to add it once I defend.

Definitely! 

If you can build GLVis yourself, the change is not hard -- the corresponding code is in the following lines:

- https://github.com/GLVis/glvis/blob/master/lib/vsdata.cpp#L500
- https://github.com/GLVis/glvis/blob/master/lib/vsdata.cpp#L512

and maybe you want a diff like:

```diff
diff --git a/lib/vsdata.cpp b/lib/vsdata.cpp
index 1934414..9e69eac 100644
--- a/lib/vsdata.cpp
+++ b/lib/vsdata.cpp
@@ -497,7 +497,7 @@ void VisualizationSceneScalarData::PrepareColorBar (double minval,
          val = ULogVal(i / 4.0);
 
          ostringstream buf;
-         buf << setprecision(4) << val;
+         buf << fixed << showpos << val;
          color_bar.addText(text_x,Y,posz, buf.str());
       }
    }
@@ -509,7 +509,7 @@ void VisualizationSceneScalarData::PrepareColorBar (double minval,
          Y = miny + (maxy - miny) * LogUVal(val);
 
          ostringstream buf;
-         buf << setprecision(4) << val;
+         buf << fixed << showpos << val;
          color_bar.addText(text_x,Y,posz, buf.str());
       }
    }
```

tzanio:
@ddkn, were you able to resolve this?

tzanio:
[...]

PR #284, opened 2024-06-19
Final changes for glvis-4.3

tzanio:
📆 Target release date: ~~July 15, 2024~~ August 7, 2024

🟡 **In Progress:**
- [ ] Update website: https://github.com/GLVis/web/pull/12

🟢 **DONE:**
- [x] Finalize and merge this PR into `master`
- [x] add test and baseline for quadrature data visualization ← @najlkin, @tzanio 
- [x] Update CHANGELOG ← @tzanio 
- [x] #302
- [x] #285 (https://github.com/GLVis/glvis/discussions/292)
- [x] #301 
- [x] #300 
- [x] #299 
- [x] #298 
- [x] #297 
- [x] #288 (see #260)
- [x] #286
- [x] Update version number from `4.2` to `4.3`
- [x] #278 ← @v-dobrev will review
- [x] Switch from `<key>A</key>` to `A` in `README.md`? See [this issue](https://github.com/github/markup/issues/1827)
- [x] #280

⚪ **After the release:**
- [ ] Update [pyglvis](https://github.com/GLVis/pyglvis/) (Merge https://github.com/GLVis/pyglvis/pull/44)
- [ ] Update [glvis-js](https://github.com/GLVis/glvis-js) and [glvis.org/live](https://glvis.org/live/)
- [ ] #296
- [ ]  #241
- [ ] Add style check to PR CIs

⚫ **Postponed:**
- #289 (see #268)
- #291
- #272
- #274
- #232 
- #170
- #166
- #133
- Support for visualization of particle data
- Adaptive geometry refiner
- Raytracing?!
- Better vector graphics output
- Improved testing with random keys

### DIFF
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2010-2022, Lawrence Livermore National Security, LLC. Produced
+// Copyright (c) 2010-2024, Lawrence Livermore National Security, LLC. Produced
 // at the Lawrence Livermore National Laboratory. All Rights reserved. See files
 // LICENSE and NOTICE for details. LLNL-CODE-443271.
 //
@@ -76,6 +76,7 @@ std::string VisualizationSceneSolution::GetHelpString() const
       << "| O -  Switch 'o' func. (NC shading) |" << endl
       << "| p/P  Cycle through color palettes  |" << endl
       << "| q -  Quits                         |" << endl
+      << "| Q -  Cycle quadrature data mode    |" << endl
       << "| r -  Reset the plot to 3D view     |" << endl
       << "| R -  Reset the plot to 2D view     |" << endl
       << "| s -  Turn on/off unit cube scaling |" << endl
@@ -84,6 +85,8 @@ std::string VisualizationSceneSolution::GetHelpString() const
       << "| y/Y  Rotate the cutting plane      |" << endl
       << "| z/Z  Move the cutting plane        |" << endl
       << "| \\ -  Set light source position     |" << endl
+      << "| Alt+a  - Axes number format        |" << endl
+      << "| Alt+c  - Colorbar number format    |" << endl
       << "| Ctrl+o - Element ordering curve    |" << endl
       << "| Ctrl+p - Print to a PDF file       |" << endl
       << "+------------------------------------+" << endl
@@ -355,7 +358,7 @@ static void KeyZPressed()
 
 static void KeyF3Pressed()
 {
-   if (vssol->shading == 2)
+   if (vssol->GetShading() == VisualizationSceneScalarData::Shading::Noncomforming)
    {
       vssol->shrink *= 0.9;
       vssol->Prepare();
@@ -369,7 +372,7 @@ static void KeyF3Pressed()
 
 static void KeyF4Pressed()
 {
-   if (vssol->shading == 2)
+   if (vssol->GetShading() == VisualizationSceneScalarData::Shading::Noncomforming)
    {
       vssol->shrink *= 1.11111111111111111111111;
       vssol->Prepare();
@@ -382,7 +385,7 @@ static void KeyF4Pressed()
 
 static void KeyF11Pressed()
 {
-   if (vssol->shading == 2)
+   if (vssol->GetShading() == VisualizationSceneScalarData::Shading::Noncomforming)
    {
       if (vssol->matc.Width() == 0)
       {
@@ -400,7 +403,7 @@ static void KeyF11Pressed()
 
 static void KeyF12Pressed()
 {
-   if (vssol->shading == 2)
+   if (vssol->GetShading() == VisualizationSceneScalarData::Shading::Noncomforming)
    {
       if (vssol->matc.Width() == 0)
       {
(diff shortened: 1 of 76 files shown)


acfisher (approved):
LGTM

najlkin:
What is the idea with quadrature functions? Interpreting them as L2 elements would be very simple :wink: . I have done that in the past in my codes.

tzanio:
> What is the idea with quadrature functions? Interpreting them as L2 elements would be very simple 😉 . I have done that in the past in my codes.

This has been discussed multiple times, but there are essentially 3 options we have considered:

1. Plot p.w. const function on the LOR Gauss-Lobatto mesh, where each LOR element contains exactly one Gauss-Legendre quadrature point. (The `p` points of 1D GLL quadrature rule are interleaved by the `p+1` points of the 1D GL rule)

2. Interpret quadrature values as the dofs of an L2 grid funcion

3. Draw a little sphere for each point

Option 3 is most general and will be required for "Support for visualization of particle data". I personally like option 1, because it does not introduce variations beyond the original data. That works only on quad/hex meshes though.

najlkin:
I see, we must reach consensus then what way to go, which might take longer than actually implementing it 😁 .

tzanio:
We can also support both options 😁

najlkin:
I found out there is no smooth vizualization in 2D for vector fields atm. This could be also added, but it would require to differentiate the vector-to-scalar function 🤔

justinlaughlin:
In the future should PRs be pulled into `glvis-4.x-dev` first instead of `master`?

tzanio:
> In the future should PRs be pulled into glvis-4.x-dev first instead of master?

No, the current policy is that PRs should go first in `master` and then we merge `master` in `glvis-4.3-dev`.

(The intention for the release branches was to be small and short-lived)

tzanio:
MFEM autotest was giving styling errors, so I run "make style" here but the results may not be fully satisfactory.

Please review https://github.com/GLVis/glvis/pull/284/commits/c697e0b465988d5a0c082edf0813705ffac75131

najlkin:
We should really add `make style` to PR CIs as #288 was merged without it 🙃

justinlaughlin:
> We should really add `make style` to PR CIs as #288 was merged without it 🙃

Oops, my bad 😅

lib/vssolution.cpp
```diff
@@ -1725,10 +1725,10 @@ void VisualizationSceneSolution::PrepareLines()
 
          MFEM_ASSERT(pointmat.Size() == 4, "Not a quadrilateral!");
 
-         //we assume that mesh_course is used only for tensor finite elements,
-         //like for representation of quadratures, so in 2D it is square
+         // we assume that mesh_course is used only for tensor finite elements,
+         // like for representation of quadratures, so in 2D it is square
          const int geom =
-            Geometry::Type::SQUARE; //ref.embeddings[i].geom; //<---- bugged!?
+            Geometry::Type::SQUARE; // ref.embeddings[i].geom; // <-- bugged!?
```
tzanio, line 1731: Do we need to leave this here?

lib/vssolution.cpp
```diff
@@ -2204,10 +2204,10 @@ void VisualizationSceneSolution::PrepareLines3()
       if (mesh_coarse)
       {
          auto &ref = mesh->GetRefinementTransforms();
-         //we assume that mesh_course is used only for tensor finite elements,
-         //like for representation of quadratures, so in 2D it is square
+         // we assume that mesh_course is used only for tensor finite elements,
+         // like for representation of quadratures, so in 2D it is square
```
tzanio, line 2208: I know this is not spelled out, but please try to leave a space after `//` or `///` when commenting for consistency with the existing code 😀

tzanio:
Should we add a test stream for quadrature function visualization?

For example [quadrature.saved.zip](https://github.com/user-attachments/files/16489744/quadrature.saved.zip) corresponding to https://github.com/GLVis/web/pull/12#issuecomment-2267634028?

lib/vssolution.cpp
najlkin: Yes or I mean without the comment maybe 😄 , but it was bugged (and it is in the referenced MFEM v4.7), see mfem/mfem#4414 .

lib/vssolution.cpp
najlkin: Okey dokey 🙃

najlkin:
> Should we add a test stream for quadrature function visualization?

That would be wise 😉 , but we do not have a matching example code for that, but I do not know if it is really a rule to have one... 🤔

tzanio:
> we do not have a matching example code for that, but I do not know if it is really a rule to have one... 🤔

it is not a hard requirement

najlkin (approved):
Everything looks good, the quadrature test would be nice if it is added, but it goes to [data](https://github.com/GLVis/data/) anyway.

tzanio:
> Everything looks good, the quadrature test would be nice if it is added, but it goes to [data](https://github.com/GLVis/data/) anyway.

I don't think it hurts to do it now. I can propose a PR with the above file, unless you prefer something else @najlkin

najlkin:
> > the quadrature test would be nice if it is added, but it goes to [data](https://github.com/GLVis/data/) anyway.
> 
> I don't think it hurts to do it now. I can propose a PR with the above file, unless you prefer something else @najlkin

No, let us do it 🚀 .

justinlaughlin:
While we have all this `glvis` momentum can we get eyes on https://github.com/GLVis/pyglvis/pull/44 so we can merge :). This can be after the `glvis` release of course. There will be other changes needed for `pyglvis` (such as updating the `glvis-js` version hosted on `npm`), but it will be nice to test with the rewrite.

PR #286, opened 2024-07-01
Visualization of quadrature functions

najlkin:
This PR adds support for visualization of quadratures (`QuadratureFunction`). They can be loaded from a file through the new command line argument `-q` or in a socket stream with the keyword `quadrature` (instead of `solution`). Two options of visualization are offered (which can be switched by `Q` key):

- Piece-wise constant function on a refined mesh (closed Gauss-Legendre refinement)
- Discontinuous elements with DOFs collocated with the quadrature on the original mesh
- L2 projection onto discontinuous elements

Note that high-order quadratures are supported only for tensor finite elements with the first two options.

TODO 📑 :

- [x] a key for switching
- [x] maintain consistency of quadrature / grid function
- [x] testing
  - [x] 1D
  - [x] vector
  - [x] stream
  - [x] animation 
  - [x] scripts
  - [x] parallel
- [x] proper refinement (ClosedGL?)
- [x] check for tensor basis
- [x] README, CHANGELOG

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -23,11 +23,12 @@ Version 4.2.1 (development)
 
 - Added visualization of quadratures. They can be loaded from a file through the
   new command line argument '-q' or in a socket stream with the keyword 'quadrature'
-  (instead of solution). Two options of visualization are offered (which can be
+  (instead of solution). Three options of visualization are offered (which can be
   switched by 'Q' key): piece-wise constant function on a refined mesh (closed
-  Gauss-Legendre refinement), or discontinuous elements with DOFs collocated with
-  the quadrature on the original mesh. High-order quadratures are supported only
-  for tensor finite elements.
+  Gauss-Legendre refinement), discontinuous elements with DOFs collocated with
+  the quadrature on the original mesh, or projection to discontinuous elements.
+  High-order quadratures are supported only for tensor finite elements with the
+  first two options.
 
 
 Version 4.2 released on May 23, 2022
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,7 +26,7 @@ Version 4.2.1 (development)
   (instead of solution). Two options of visualization are offered (which can be
   switched by 'Q' key): piece-wise constant function on a refined mesh (closed
   Gauss-Legendre refinement), or discontinuous elements with DOFs collocated with
-  the quadrature on the original mesh.
+  the quadrature on the original mesh (available for tensor basis only).
 
 
 Version 4.2 released on May 23, 2022
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -26,7 +26,8 @@ Version 4.2.1 (development)
   (instead of solution). Two options of visualization are offered (which can be
   switched by 'Q' key): piece-wise constant function on a refined mesh (closed
   Gauss-Legendre refinement), or discontinuous elements with DOFs collocated with
-  the quadrature on the original mesh (available for tensor basis only).
+  the quadrature on the original mesh. High-order quadratures are supported only
+  for tensor finite elements.
 
 
 Version 4.2 released on May 23, 2022
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -25,14 +25,13 @@ Version 4.2.1 (development)
 
 - Added a compilation parameter for the default font size.
 
-- Added visualization of quadrature data (QuadratureFunction in MFEM). They can
-  be loaded from a file through the new command line argument '-q' or in a socket
-  stream with the keyword 'quadrature' (instead of 'solution'). Three options of
-  visualization are offered (which can be switched by 'Q' key): piece-wise constant
-  function on a refined mesh (closed Gauss-Legendre refinement), discontinuous
-  elements with DOFs collocated with the quadrature on the original mesh, or
-  projection to discontinuous elements. High-order quadrature functions are supported
-  only for tensor finite elements with the first two options.
+- Added visualization of quadrature data (QuadratureFunction in MFEM). Both
+  loading from file, with the new command line argument '-q', or from a socket
+  stream, with the keyword 'quadrature', are supported. Three visualization
+  options are provided: piece-wise constants on a refined mesh (LOR), L2 field
+  with DOFs collocated (interpolation), or projection to discontinuous elements
+  (L2 projection). Use 'Q' to switch between them. High-order quadrature data is
+  supported only for tensor finite elements with the first two options.
 
 
 Version 4.2 released on May 23, 2022
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -21,6 +21,13 @@ Version 4.2.1 (development)
 
 - Added a compilation parameter for the default font size.
 
+- Added visualization of quadratures. They can be loaded from a file through the
+  new command line argument '-q' or in a socket stream with the keyword 'quadrature'
+  (instead of solution). Two options of visualization are offered (which can be
+  switched by 'Q' key): piece-wise constant function on a refined mesh (closed
+  Gauss-Legendre refinement), or discontinuous elements with DOFs collocated with
+  the quadrature on the original mesh.
+
 
 Version 4.2 released on May 23, 2022
 ====================================
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,7 +31,9 @@ Version 4.2.1 (development)
   options are provided: piece-wise constants on a refined mesh (LOR), L2 field
   with DOFs collocated (interpolation), or projection to discontinuous elements
   (L2 projection). Use 'Q' to switch between them. High-order quadrature data is
-  supported only for tensor finite elements with the first two options.
+  supported only for tensor finite elements with the first two options. With the
+  first option, only the mesh lines of the original mesh are visualized. This
+  feature is also supported for the element-wise cutting plane (cplane=2).
 
 
 Version 4.2 released on May 23, 2022
--- a/README.md
+++ b/README.md
@@ -145,9 +145,9 @@ Key commands
   vector formats (SVG, EPS) are also possible, but keep in mind that the
   printing takes a while and the generated files are big.
 - <kbd>Q</kbd> – Cycle between representations of the visualized *quadrature data*. The options are:
-  - piece-wise constant refined
-  - L2 element dof collocation
-  - L2 element projection
+  - piece-wise constant refined (LOR)
+  - L2 element dof collocation (interpolation)
+  - L2 element projection (L2 projection)
 - <kbd>q</kbd> – Exit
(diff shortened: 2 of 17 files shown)


najlkin:
Unfortunately, there is nothing for high-order quadratures on non-tensor elements. The number of DOFs matches for L2 elements, but the points are different. We would need some special basis for that in MFEM 🤔 .

v-dobrev:
> Unfortunately, there is nothing for high-order quadratures on non-tensor elements. The number of DOFs matches for L2 elements, but the points are different. We would need some special basis for that in MFEM 🤔 .

Using the values at the quadrature points, one can compute a r.h.s vector for a local L2 projection on an appropriate L2 basis. There are probably multiple options for selecting the "appropriate" L2 basis. One possibility is to choose the order of the L2 basis based on the order of the quadrature rule -- e.g. if the quadrature rule is exact for polynomials of degree p, then we project on order p/2 basis -- this will ensure that we get invertible mass matrix if we use the same quadrature rule for the mass matrix as for the r.h.s.

najlkin:
Well, that would work for sure, but there many different ways, as you say, so it is not less of visualization and more of processing. It might introduce some oscillations and does not necessarily match the quadrature point-wise. Therefore, I am not sure we want that 🤔

najlkin:
Ok, you convinced me @v-dobrev , that was quick 😄 . I added L2 projection is another option, which is also the fallback for the other two when non-tensor elements are used. The order is the same as with that collocation, so the same number of points, which seems to me as the most logical. As the quadrature points and DOFs are close, the resulting grid function looks fine :wink: .

README.md
```diff
@@ -144,7 +144,8 @@ Key commands
 - <kbd>Ctrl</kbd> + <kbd>p</kbd> – Print to a PDF file using `gl2ps`. Other
   vector formats (SVG, EPS) are also possible, but keep in mind that the
   printing takes a while and the generated files are big.
-- <kbd>Q</kbd> – Cycle between representations of the quadrature (piece-wise constant refined/L2 element dof collocation) 
+- <kbd>Q</kbd> – Cycle between representations of the quadrature (piece-wise constant refined
+  / L2 element dof collocation / L2 element projection)
```
v-dobrev, line 148: Wasn't `Q` supposed to close all GLVis windows? Or was that just an idea that never got implemented? @tzanio, do you remember?

tzanio: > Wasn't `Q` supposed to close all GLVis windows? Or was that just an idea that never got implemented? @tzanio, do you remember?

I don't think such a functionality was implemented.

tzanio:
I think this is really awesome, thank you for adding it @najlkin! 👍 

To speed-up my review: do you have a simple driver to define and visualize a quadrature function and/or sample quadrature function files?

README.md
```diff
@@ -144,6 +144,8 @@ Key commands
 - <kbd>Ctrl</kbd> + <kbd>p</kbd> – Print to a PDF file using `gl2ps`. Other
   vector formats (SVG, EPS) are also possible, but keep in mind that the
   printing takes a while and the generated files are big.
+- <kbd>Q</kbd> – Cycle between representations of the quadrature (piece-wise constant refined
```
tzanio, line 147: See "quadrature" comment above

najlkin:
> To speed-up my review: do you have a simple driver to define and visualize a quadrature function and/or sample quadrature function files?

In `ex1.cpp`, add: 
```
QuadratureSpace qspace(&mesh, order);
QuadratureFunction qf(&qspace);
qf.ProjectGridFunction(x);
ofstream qf_ofs("sol.qf");
qf_ofs.precision(8);
qf.Save(qf_ofs);
```
and then in the visualization part instead of `...solution...` use: `sol_sock << "quadrature\n" << mesh << qf << flush;` 👍

CHANGELOG
```diff
@@ -21,6 +21,15 @@ Version 4.2.1 (development)
 
 - Added a compilation parameter for the default font size.
 
+- Added visualization of quadratures. They can be loaded from a file through the
```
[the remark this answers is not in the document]
najlkin: Done :wink:

tzanio:
- [ ] Add description for `Q` in the help messages printed by `h`

lib/stream_reader.cpp
```diff
@@ -115,22 +211,134 @@ void StreamState::SetMeshSolution()
...
+         internal.mesh.reset(mesh_lor);
+      }
+      break;
+      case QuadSolution::HO_L2_collocated:
+      {
+         FiniteElementCollection *fec = new L2_FECollection(order, mesh->Dimension());
+         FiniteElementSpace *fes = new FiniteElementSpace(mesh.get(), fec,
+                                                          quad_f->GetVDim(), Ordering::byVDIM);
+         MFEM_ASSERT(quad_f->Size() == fes->GetVSize(), "Size mismatch");
+         cout << "Representing quadrature by grid function of order "
```
tzanio, line 272: Here and below "quadrature data"

najlkin:
We have a bigger problem here, it cannot pass on Mac suddenly after just changing README and CHANGELOG 🙃

lib/stream_reader.cpp
najlkin: Done 👍

tzanio:
This is super cool!

Here are the three quadrature data visualization modes for Example 1p on `star-q3.mesh`

| LOR             |  Interpolation |  Projection |
:-------------------------:|:-------------------------:|:-------------------------:
Screenshot 2024-07-17 at 11 58 51 AM | Screenshot 2024-07-17 at 11 58 54 AM | Screenshot 2024-07-17 at 11 58 58 AM

tzanio:
> We have a bigger problem here, it cannot pass on Mac suddenly after just changing README and CHANGELOG 🙃

The CI is quirky, don't worry about it

tzanio:
Works in 3D, parallel, etc. too 😁 

Screenshot 2024-07-17 at 12 13 08 PM

vladotomov:
For the LOR case, is there some way to visualize mesh faces differently than internal LOR faces? Maybe draw only the mesh faces?

tzanio (approved):
I added some small changes, but I think this is pretty awesome as it is, thanks @najlkin! 🚀 

As a follow-up I think we should add use of this in appropriate places in MFEM, e.g. in the Mesh Explorer miniapp.

tzanio:
> For the LOR case, is there some way to visualize mesh faces differently than internal LOR faces? Maybe draw only the mesh faces?

I agree that this will be nice, but I'm not sure it is feasible.

As an alternative, we can give the same element attribute to the LOR elements from the same parent and pull them apart in GLVis

najlkin:
> For the LOR case, is there some way to visualize mesh faces differently than internal LOR faces? Maybe draw only the mesh faces?

@vladotomov  How it works internally is that it converts the quadrature function to grid function before the visualization happens, so there is no way how to do that currently, but it is true I store the original mesh (non-refined), so we could theoretically tunnel the original mesh to the visualization code for the mesh lines one day 😉 But I would not delay this PR and release because of that.

tzanio:
> > We have a bigger problem here, it cannot pass on Mac suddenly after just changing README and CHANGELOG 🙃
> 
> The CI is quirky, don't worry about it

If you look at the failing Mac screenshots at the bottom of https://github.com/GLVis/glvis/actions/runs/9980266401?pr=286, the zoom level appear a bit different. Maybe something changes in GitHub, but just to double-check: are we sure nothing changed with respect to zoom or keys handling that could cause this?

test fail shifted saved

najlkin:
I do not believe so, nothing of that sort. Mainly, it was passing, I just changed README and CHANGELOG 😂

tzanio:
It could be a change in GitHub's mac version or something.

Just to triple-check, can you try some of the tests from https://github.com/GLVis/data/tree/53fbee222cceb0581122cb72518bfc1b85766b4e with master and this PR and make sure the images are the same size?

najlkin:
I do not have Mac and on Ubuntu it passes.

tzanio:
> I do not have a Mac and on Ubuntu it passes.

Good point 😄 

I manually re-run all the test in [this PR](https://github.com/GLVis/glvis/actions/runs/9980266401) and in [#483](https://github.com/GLVis/glvis/actions/runs/9966900943) (which passed before)... Let's see what happens

najlkin:
And now someone tell me how a text file can break a code on Mac and only with CMake 😂

najlkin:
Btw, I am close to implementing that @vladotomov suggestion. The difficult part was not the tunneling the mesh, but mixing the meshes together in 2D, where the surface is lifted by the values on the refined mesh, but I have solved that. So like in an hour I may have it 😉 .

tzanio:
Looks like https://github.com/GLVis/glvis/pull/286/commits/f25e99eb03776702991b9d88ac95cb44a1282c61 fixed the CI

tzanio:
Here is the above picture with the mesh drawn as @vladotomov suggested

Screenshot 2024-07-18 at 4 45 54 PM

tzanio:
I like the latest changes @najlkin, re-approving 👍

najlkin:
Wait for 3D 😮 😎

justinlaughlin:
image

Looks very cool! Just curious, is there a way to toggle back and forth between quadrature function / grid function or is it one or the other? I naively tried doing both but it didn't like that.

```bash
./glvis -np 8 -m ../mfem-parallel/examples/mesh -g ../mfem-parallel/examples/sol -q ../mfem-parallel/examples/sol_qf
```

I see a lot of nice refactoring work here too like with the `enum` :).

justinlaughlin (approved):

najlkin:
Nice to hear @justinlaughlin , you need to choose what you want to visualize - a grid function or quadrature function. If you have a bit of time, please test different options, like dimensions, vector/scalar, shadings, ...

najlkin:
Ok, I think I am done 🙌 . There is 3D scalar/vector visualization of original quad lines, including element-wise cutting plane (`cplane=2`) 😉 Flat cutting plane (`cplane=1`) would be difficult and I will leave it for some other day.

justinlaughlin:
> Nice to hear @justinlaughlin , you need to choose what you want to visualize - a grid function _or_ quadrature function. If you have a bit of time, please test different options, like dimensions, vector/scalar, shadings, ...

Forgot to mention I tried some things using ex1: different meshes, mesh refinement, function space order, serial/parallel, shadings, stream/command line; I saw no issues. I can test it out more but it would be next week. Looks good to me!

justinlaughlin:
oh, forgot to mention I tested on WSL2 ubuntu 20.04

tzanio:
Very cool, thanks @najlkin!!

Screenshot 2024-07-19 at 1 07 25 PM

PR #289, opened 2024-07-11
Color palettes

justinlaughlin:
## Refactor Palettes for better customization

This PR:
 - Provides a more flexible implementation for color palettes, primarily to allow specification of palettes both at compile and run time.
 - Compile time palettes are defined under `lib/base_palettes.cpp` and exported as `BasePalettes`. Some of these were generated by functions and are now just defined explicitly.
 - Run time palettes can be specified using the command line argument `-p <palette_filename>`; an example palette file with the proper structure is under `share/palettes.txt` (this includes all of the palettes mentioned in the old version of the PR).
 - Adds a few structs/classes
   - Struct `RGBAf` (to represent RGBA in [0,1] float format)
   - Struct `Palette` (holds `vector<RGBAf> colors` and `string name`)
   - Class `PaletteRegistry` (holds `vector<shared_ptr<Palette>>`)
 - Updates `PaletteState`; previously it required the global variables `Num_RGB_Palettes`, `RGB_Palette_Sizes`, `RGB_Palettes`, and `RGB_Palette_Names`. Now it uses a pointer to a `PaletteRegistry`. By default, a `PaletteState` will be initialized with the global variable `BasePalettes` but it is not necessary.

Misc notes:
 - Palettes can be loaded with explicit alpha channels now; right now this doesn't do anything. I think this would require some changes to `PaletteState::GenerateAlphaTexture`.
 - Right now because `BasePalettes` is updated before any `PaletteState` are created, they initialize properly, but because textures are created by `PaletteState::Init` this could lead to problems if palettes are loaded in at a later point in the run. Moving the functionality to `Palette`, eg `Palette::as_texture()` with caching could be a solution. (there are several other methods that should be moved to `Palette` so that `PaletteState` is just in charge of the state).
 - Palettes can now be retrieved by name but this isn't used anywhere yet.

## Original PR

There is a nice resource from Fabio Crameri that provides a lot of color-vision deficiency friendly and perceptually-uniform color maps [see [issue 268](https://github.com/GLVis/glvis/issues/268), and [reference](https://zenodo.org/records/8409685)]. This PR adds most of those colormaps into GLVis.

Here is an example of `batlow`

I wrote a small script to help munge the colormap files into something copy-pasteable. I'm not sure if it's worth adding to the repo but I posted it as a [gist](https://gist.github.com/justinlaughlin/2d430baf401fbbb6a2f1719005a5139c). The colormaps added in this PR are listed in the gist:

```python
cmnames = ['batlow', 'batlowW', 'batlowK', 'glasgow', 'lipari', 'navia', # continuous
           'oleron', 'bukavu', 'fes', # multi-sequential
           'hawaii', 'buda', 'imola', # discrete
           'oslo', 'nuuk', 'lajolla', 'bamako', 'davos', # categorical
           'bilbao', 'lapaz', 'acton', 'turku', 'tokyo',
           'broc', 'cork', 'vik', 'lisbon', 'tofino', 'berlin', # diverging
           ]
```


justinlaughlin:
It looks like tests 15 and 17 are failing because they cycle color palettes backwards using `P`, so they get around to the new palettes. Maybe we should restrict palettes that cycle to a subset of the total? Or update tests?

v-dobrev:
This seems like going a bit overboard with the palettes. 😁  The size of `palettes.cpp` is essentially doubled in terms of lines!

justinlaughlin:
I started with just a few, but it was taking a while so I wrote a script to help, and then I figured might as well add most 😆. I think we had 43 before and this adds 28? Could pare it down too...

justinlaughlin:
Some suggestions from gm:
 - Add option for user to point to custom colormap file
 - Keep "cyclable" colormaps the same for backwards compatability

Extra colormaps + user defined ones would still be accessed through F6?

tzanio:
To clarify

> Add option for user to point to custom colormap file

The idea will be to have an optional colormap file (could be configurable at build time, or just "colormaps.dat" in the same directory as the executable) and for `glvis` to check for that file at startup and if found to add the additional colormaps at the end of the default list.

This way, we preserve the default behavior, while allowing for local customization.

tzanio:
@justinlaughlin, I think it will be still cool to add an optional input file for this. Happy to chat if you want to give it a try.

justinlaughlin:
> @justinlaughlin, I think it will be still cool to add an optional input file for this. Happy to chat if you want to give it a try.

@tzanio I pushed some changes and updated the PR description. So far it works for me; feel free to give it a try with `-p shared/palettes.txt`

lib/palettes.cpp
```diff
@@ -7703,16 +221,15 @@ void PaletteState::Init()
       MaxTextureSize = std::min(MaxTextureSize, 4096);
       {
          std::lock_guard<std::mutex> lk{init_mtx};
-         Init_Palettes();
       }
 
-      GLuint paletteTexIds[Num_RGB_Palettes][2];
+      vector<array<GLuint, 2>> paletteTexIds(Palettes->MAX_PALETTES);
```
najlkin, line 226: Why this? 🤪  The number of palettes cannot be changed after init anyway, because there is no re-init...

lib/palettes.hpp
```diff
@@ -55,7 +56,11 @@ class PaletteState
    { return palette_tex[curr_palette][use_smooth]; }
    GLuint GetAlphaTexture() const { return alpha_tex; }
    void GenerateAlphaTexture(float matAlpha, float matAlphaCenter);
+
+   int NumPalettes();
```
najlkin, line 60: Conversion from `size_t` to `int`. I would choose one consistently.

najlkin (commented):
Thanks for cleaning this up in GLVis, but needs some polishing🙂 .

najlkin:
Also I would separate the classes from the color palette arrays and split declaration / definition to make the code easier to navigate, so maybe move the classes to a new file or put them to `palette.cpp/hpp` at least 😉 .

tzanio:
- [ ] Please, describe in `CHANGELOG`

share/palettes.txt
```diff
@@ -0,0 +1,7224 @@
+palette batlow RGBf
```
tzanio, line 1: Consider adding comments to indicate the source of the palettes (https://zenodo.org/records/8409685?) and/or changing the name of the file, e.g. to `palettes-crameri.txt`.

We should probably also mention the license at the end of our `NOTICES` file?

In the future, we could potentially have different `palettes-*.txt` files in `share/` from different sources.

tzanio (approved):
This is great and exactly what I was hoping for, thanks @justinlaughlin!

Approved modulus small comments and @najlkin's suggestions.

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+      for (const auto& color : colors)
+      {
+         color.print(false, os);
+         os << endl;
+      }
+      os << endl;
+   }
+
+   // helper function - not used in glvis.cpp
+   void printAsCPP(const string& filename, const string& varname) const
```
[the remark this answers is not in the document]
justinlaughlin: I used this to help format the previous palettes. I'm okay with removing it.

lib/palettes.cpp
justinlaughlin: Initialization of `paletteTexIds` using `NumPalettes()` throws a CI error because the dynamic allocation isn't compatible with `c++99`. There is no re-init atm but I think most (all?) of `PaletteState::Init()` should go into `Palette` anyways; that way you can add palettes at any point, eg in `pyglvis`.

najlkin: I understand that a C array is a problem, but `std::vector` should be ok. Secondly, where is `c++99`? MFEM requires C++ 11 🤪 .
It is batched here in a way, but it should not be a big performance hit to allocate the textures separately in `Palette` directly, I believe, and it would be much cleaner 😉 .

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+      outfile << endl;
+      outfile.close();
+   }
+
+   shared_ptr<Palette> shared() const
+   {
+      return make_shared<Palette>(*this);
+   }
+
+   double* as_rgb_array() const
```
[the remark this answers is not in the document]
justinlaughlin: This was a mistake - added const back.

lib/palettes.hpp
```diff
@@ -35,9 +36,9 @@ class PaletteState
    int ChoosePalette();
    int SelectNewRGBPalette();
    /// Gets the data in the palette color array.
-   const double* GetData() const;
```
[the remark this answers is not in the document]
justinlaughlin: mistake - added const back

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+   {
+      // palette name is unique || container is empty
+      if (get_index_by_name(name) == -1 || palettes.empty())
+      {
+         addPalette(Palette(name));
+      }
+   }
+
+   // get by index
+   shared_ptr<Palette> get(size_t index) const
```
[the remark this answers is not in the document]
justinlaughlin: Changed to unique_ptr

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+   }
+};
+
+
+// PaletteRegistry with a vector of shared pointers to Palette
+// Besides holding the palettes, this should be stateless
+class PaletteRegistry
+{
+private:
+   vector<shared_ptr<Palette>> palettes;
```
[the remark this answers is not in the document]
justinlaughlin: changed throughout to use `unique_ptr`

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+   // --- Constructors ---
+   Palette(const string& name) : name(name) {}
+   Palette(const string& name, const vector<RGBAf>& colors)
+      : name(name), colors(colors) {}
+
+   // from Nx3 array
+   template <size_t N>
+   // Palette(const string& name, const float (&array)[N][3]) : name(name) {
+   Palette(const string& name, const array<array<float,3>,N>& arr) : name(name)
+   {
```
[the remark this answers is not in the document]
justinlaughlin: good idea, added it in.

lib/base_palettes.hpp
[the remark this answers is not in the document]
justinlaughlin: removed.

lib/palettes.hpp
```diff
@@ -55,7 +56,11 @@ class PaletteState
    { return palette_tex[curr_palette][use_smooth]; }
    GLuint GetAlphaTexture() const { return alpha_tex; }
    void GenerateAlphaTexture(float matAlpha, float matAlphaCenter);
+
+   int NumPalettes();
 private:
+   bool first_init;
+   PaletteRegistry* Palettes;
    void ToTextureDiscrete(double * palette, size_t plt_size, GLuint tex);
    void ToTextureSmooth(double * palette, size_t plt_size, GLuint tex);
```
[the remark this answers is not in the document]
justinlaughlin: good idea - I updated each to take a `Palette*`

share/palettes.txt
justinlaughlin: added in [190e5c4](https://github.com/GLVis/glvis/pull/289/commits/190e5c40e4d5212e6819da5214ef4164a2cc8c4d)

lib/palettes.cpp
justinlaughlin: https://github.com/GLVis/glvis/actions/runs/11693833399/job/32566220453#step:17:27 is what I was referring to, but looking back it was an array not a vector back then 🤔 maybe I'm misremembering 😅. 

Regardless, yeah I'll just try to move this and other relevant things into `Palette`; it'll be much cleaner and would also be a nice way to utilize the alpha channel in the palette file. The only thing I'm not sure of is currently the alpha adjustment <kbd>k</kbd> / <kbd>K</kbd> is a global operation. If a colormap has a custom alpha channel and the user makes an adjustment, do we override/ignore/multiply?

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+         palette->print(os);
+      }
+   }
+
+   size_t NumPalettes() const
+   {
+      return palettes.size();
+   }
+
+   int get_index_by_name(const string& name) const
```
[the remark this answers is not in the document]
justinlaughlin: done

lib/palettes.hpp
justinlaughlin: changed to `int` throughout. Some of the `PaletteState` methods like `GetSize` use `-1` as a check.

lib/palettes.cpp
najlkin: Yeah, it is about a C array, where you can use only `constexpr`. It just mentions variable-length arrays from C99 as a suggestion, but that is a different story.
That is a clean approach to move things to `Palette`. With alpha, I would multiply, that is the most consistent behavior, I would say 🤔 .

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+// PaletteRegistry with a vector of shared pointers to Palette
+// Besides holding the palettes, this should be stateless
+class PaletteRegistry
+{
+private:
+   vector<shared_ptr<Palette>> palettes;
+
+public:
+   const static size_t MAX_PALETTES = 1000;
```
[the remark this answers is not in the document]
najlkin: ☝️  Remove please.

lib/palettes.cpp
```diff
@@ -7615,28 +133,30 @@ void PaletteState::ToTextureDiscrete(double * palette, size_t plt_size,
 }
 
 // Generates a smooth texture from the given palette.
-void PaletteState::ToTextureSmooth(double * palette, size_t plt_size,
-                                   GLuint tex)
+void PaletteState::ToTextureSmooth(Palette* palette, GLuint tex)
 {
+   int plt_size = palette->size();
+   const double * paldata = palette->as_rgb_array();
```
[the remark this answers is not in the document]
justinlaughlin: Fixed, I think, in e571cd0

lib/base_palettes.hpp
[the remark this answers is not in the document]
justinlaughlin: done

lib/palettes.cpp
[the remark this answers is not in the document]
najlkin: Looks good, but maybe you may drop `cache_updated` and use `colors.size() == rgb_cache.size()`. The less variables the less error prone 😉 .

lib/base_palettes.hpp
```diff
@@ -147,7 +147,7 @@ class Palette
 // Since there is no make_unique for c++11
 #if __cplusplus < 201402
```
najlkin, line 148: And when it is higher, `as_unique` will not be defined 😅

justinlaughlin: oops haha, thank you

justinlaughlin:
I noticed that discrete textures don't use `RepeatPaletteTimes` besides for reversing (so you need to press <kbd>!</kbd> to change to the smooth version), is this intended?

tzanio:
> you need to press ! to change to the smooth version

If I understand correctly, that is intended, so you can get a visualization in several bands:

Screenshot 2024-11-14 at 12 43 41 PM

najlkin:
I think @justinlaughlin means that the discrete palette does not repeat even when you specify that, while the smooth palette does. That looks inconsistent to me 🤔 .

justinlaughlin:
> If I understand correctly, that is intended, so you can get a visualization in several bands:

I guess what I mean is when I press <kbd>F6</kbd> it asks for 
1) Palette repetition times ($R$)
2) Palette index
3) Number of colors ($N$)

If I'm in discrete mode and set $R=3$ and $N=5$ I'd expect to get 15 colors (in same ordering used by the smooth palette) but it is just 5. 

```
Palette is repeated 1 times.
(Negative value means the palette is flipped.)
Enter new value: 3
Palette will be repeated 3 times now.

Choose a palette:
...
 ---> [1]

Palette is using 5 colors.
Enter new value (0 = use original 5 colors): 0
Palette will be using 5 colors now.
```

Discrete
image

Smooth
image

najlkin:
Would it be difficult to fix, @justinlaughlin ? ... if we agree it is to be fixed, but I believe we do 😅

lib/palettes.cpp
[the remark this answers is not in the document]
justinlaughlin: based on how this was being used, I changed it to just return a copy.

justinlaughlin:
> Would it be difficult to fix, @justinlaughlin ? ... if we agree it is to be fixed, but I believe we do 😅

I changed this and simplified the interpolation logic a bit in the `Texture` class. As a nice side-effect we can use custom alpha channels (seems to work better with a black background); turns out the multiplication is already happening, I think because of `GL_BLEND`?

image

lib/base_palettes.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+};
+
+
+struct Palette
+{
+   const string name;
+   vector<RGBAf> colors;
+
+   // --- Constructors ---
```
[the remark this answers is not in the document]
justinlaughlin: I'll add some more comments; saving the most fun part for last 😄

lib/palettes.cpp
justinlaughlin: I think we can resolve this?

lib/openglvis.hpp
```diff
@@ -195,7 +195,8 @@ class VisualizationScene
    glm::mat4 rotmat;
    glm::mat4 translmat;
 
-   float matAlpha = 1.0;
+   // Needs to be < 1.0 to allow alpha defined at palette level to show through
+   float matAlpha = 0.999999;
```
najlkin, line 199: Why this? I thought they are multiplied 🤔

tzanio:
The transparent colormaps are cool! Do they work ok for data on 3D meshes?

lib/openglvis.hpp
justinlaughlin: I don't understand it enough yet to explain why 😅 . When it is 1 though, all the alpha values set by the discrete/smooth texture for each palette are ignored.

justinlaughlin:
> The transparent colormaps are cool! Do they work ok for data on 3D meshes?

image

Doesn't look like it 😞 ; it works on cutting planes though 🤔. I need to dig a bit more to understand what `GL` is doing.

lib/openglvis.hpp
justinlaughlin: There is a check `param.is_translucent = matAlpha < 1.0` in several areas. I added a check in [006a042](https://github.com/GLVis/glvis/pull/289/commits/006a042df35e21ad105daa22bc539a77f14da90c) so it shouldn't be necessary anymore.

justinlaughlin:
Turns out, it does actually work on 3d meshes (I don't think we have internal boundaries? so just the external mesh). The example palettes I used just happened to have $\alpha=0$ at the lowest end of the scale, and I was testing it on `mfem` `ex1` solutions using zero dirichlet bcs 😅.

In the future, a volume rendering mode would be very nice.

justinlaughlin:
Okay, I think this is ready to go

CHANGELOG
```diff
@@ -9,6 +9,20 @@
                              https://glvis.org
 
 
+Version 4.y.z released on
```
tzanio, line 12:
```suggestion
Version 4.3.3 (development)
```

tzanio (changes requested):
Approved modulo the renaming suggestions I put in the comments.

We can also consider adding some additional palettes, e.g.

- The [ColorCET](https://colorcet.com/,) perceptually uniform colour maps from [here](https://github.com/GLVis/glvis/blob/fb40757c7326e9cac7286c3f6d2bbd77e78c33f6/lib/cet_palettes.hpp).
- The VisIt's colortables from [here](https://github.com/visit-dav/visit/tree/develop/src/resources/colortables).
- The MatPlotLib colormaps, see [here](https://matplotlib.org/stable/users/explain/colors/colormaps.html).

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+public:
+   /// Empty constructor
+   PaletteRegistry() {}
+
+   /// Constructor via a const vector of Palettes; if name already exists, skip
+   /// Used for loading compiled palettes (i.e. `palette_definitions.cpp`)
+   PaletteRegistry(const vector<Palette>& paletteRefs);
+
+   /// Adds an existing palette to the registry
+   void addPalette(Palette& palette);
```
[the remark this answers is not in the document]
justinlaughlin: Updated except for `get()` and `size()` which seem to already use lowercase elsewhere.

lib/palette_definitions.cpp
```diff
@@ -0,0 +1,7575 @@
+// Copyright (c) 2010-2024, Lawrence Livermore National Security, LLC. Produced
```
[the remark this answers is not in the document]
justinlaughlin: Changed to `palettes_default.cpp`

lib/palettes_base.hpp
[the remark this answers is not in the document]
najlkin: Like `Array::Size()`? 🙂 I also do not insist, but when renaming, then all, I would say.

justinlaughlin:
> I have a point for discussion here. Alpha channel is now supported by palettes technically, but except two special external palettes there is no way how to use this feature. So how about adding some option to add alpha channel to existing palettes? For example, there could be an extra option after pressing F6, which would add a simple gradient from 0 to 1 to your selected palette and you could optionally set the repetition/reversion in a similar way to the (non-alpha) color palette? 🤔 That would work nicely in most cases without the need to have special palettes for that 😉 . Technically, we could have two classes for palettes with one base, where one would be for RGB only palettes and the second one for RGBA palettes instead of defaulting the alpha channel during loading. This extra option would be offered only for the former one.

I like your suggestion - but I think it would make more sense as an extension of the existing "global" alpha modifier (ie the modifier changed by<kbd>k</kbd>, <kbd>K</kbd>, <kbd>,</kbd>, <kbd><</kbd>) rather than the palette alpha channel itself. Palette repetition/reversion is already a global state (represented in `PaletteState`; all `Texture`s are regenerated when palette repetition/reversion is changed).

IMO the main use for custom alpha maps is to highlight certain parts of the data (kind of like more customizable level sets), especially when there is lots of occlusion eg in volumetric rendering. The [NVIDIA IndeX](https://www.nvidia.com/en-us/data-center/index-paraview-plugin/) plugin has some nice examples. But, in order to do that you need a more sophisticated alpha editor (paraview lets you define it as a piecewise linear function). I think what you want to highlight is often pretty specific to the data/palette.

TBH I think the global modifier is already good as-is; it could get kind of complicated if we add too many features (what if you reverse the palette, but don't want the alpha channel reversed too?). I think some useful directions for the future may be 1) more formats to load from (paraview exports palettes as a `json` iirc), 2) a piecewise editor, 3) support in `pyglvis` (it would be quite easy to modify your palette as a `numpy` array).

tzanio (approved):

najlkin:
Sure, some easier ways how to define palettes would be nice and may include that alphas 😉 . But I was thinking of something basic at the level of the default palettes, where you also cannot specify any levels you want highlight, but still most of the people are fine with that. Most of the palettes are linear, highlighting one end of the scale, so that would match the 0 to 1 gradient, or they have uniformly distributed levels, which can be achieved by repetition of the alpha gradient.
I agree that global solution makes more sense, I just mean that I would not try to mix it with the alpha channel of the palette and only add that when it is absent, so the logic may stay relatively simple.

justinlaughlin:
> Approved modulo the renaming suggestions I put in the comments.
> 
> We can also consider adding some additional palettes, e.g.
> 
> * The [ColorCET](https://colorcet.com/,) perceptually uniform colour maps from [here](https://github.com/GLVis/glvis/blob/fb40757c7326e9cac7286c3f6d2bbd77e78c33f6/lib/cet_palettes.hpp).
> * The VisIt's colortables from [here](https://github.com/visit-dav/visit/tree/develop/src/resources/colortables).
> * The MatPlotLib colormaps, see [here](https://matplotlib.org/stable/users/explain/colors/colormaps.html).

I added the cet maps `share/palettes-cet.txt`.

For VisIT maybe in a future PR we should add a loader for their `.ct` specification? Looking at it briefly it looks like they allow for custom spacing of points and also different versions of the spec. If we need to parse it we might as well integrate that into GLVis.

I'd need to dig a bit more on matplotlib's license (they say it is a mixture of the licenses of each contributor?); they do have a nice list of [externally maintained colormaps](https://matplotlib.org/thirdpartypackages/#colormaps-and-styles) that maybe we can consider adding from in the future?

justinlaughlin:
> Technically, we could have two classes for palettes with one base, where one would be for RGB only palettes and the second one for RGBA palettes instead of defaulting the alpha channel during loading.

We could also use the check `Palette::IsTranslucent()` which checks if any `alpha != 1.0`. 

> Most of the palettes are linear, highlighting one end of the scale, so that would match the 0 to 1 gradient, or they have uniformly distributed levels, which can be achieved by repetition of the alpha gradient.

Now that I think about it more, I'd prefer a less context dependent solution. What if a user uploads a palette with no alpha but wants to add a linear mask? I think the issue is that <kbd>F6</kbd> already asks for a lot of information at once as is; I'd personally prefer if we implemented something like https://github.com/GLVis/glvis/issues/287 first, and then add this alpha adjustment idea as a command. 

I suggest we separate this into a different PR. Many GLVis commands are already kind of confusing (IMO 🫣) and I think this would add more confusion. E.g. <kbd>k</kbd>, <kbd>K</kbd> sets one kind of alpha mask (I assumed it was uniform, but once I read the source I realized its a two-sided exponential curve (see plot); with a peak shifted by <kbd>,</kbd> and <kbd><</kbd>). If <kbd>F6</kbd> is (sometimes) used to set a linear mask (and another key else for uniform? something else for piecewise linear?) that is too many different keys + context to keep track of. 

image

najlkin:
Wow 😳 , I ...and probably all other users 😄 ... did not know it is complicated like that 🙃 . But when you look at that, it is actually close to what I was proposing with the linear gradient. So definitely I would not add another layer of logic to the alpha channel, but it is another reason to implement the logic in palettes to unify the whole thing. Also incorporating the settings to the `F6` menu would give better control over it to the user. Anyway, we may defer it to another PR, this is good as it is 😉

lib/palettes_base.cpp
```diff
@@ -0,0 +1,337 @@
...
+   // Generate the smooth texture data (interpolates colors)
+   else
+   {
+      for (int rpt = 0; rpt < Nrepeat; rpt++)
+      {
+         bool reverse = (reversed + rpt) % 2 != 0;
+         for (int i = 0; i < Ncolors; i++)
+         {
+            float t = 0.999999 * i * (plt_size - 1) / (Ncolors - 1);
+            int j = floor(t);
```
najlkin, line 160:
```suggestion
            float t = i * (plt_size - 1) / (Ncolors - 1);
            int j = std::min((int)t, plt_size - 1);
```

lib/palettes_base.cpp
```diff
@@ -0,0 +1,337 @@
...
+         if (idx == -1)
+         {
+            AddPalette(palname);
+            idx = GetIndexByName(palname);
+            cout << "Reading palette: (" << idx+1 << ") " << palname << endl;
+         }
+         else
+         {
+            cout << "Error reading palette: " << palname
+                 << ". Palette with same name already exists." << endl;
```
najlkin, line 312: I do not know, does this need to be an error stopping the loading? Maybe you have two large collections, which partially overlap, and you do not want to go through them manually to remove duplicates. I would prefer a warning and skipping or rewriting the palette.

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   /// Constructor from Nx3 array
+   template <size_t N>
+   Palette(const string& name, const array<array<float,3>,N>& arr);
+
+   /// Constructor from Nx4 array
+   template <size_t N>
+   Palette(const string& name, const array<array<float,4>,N>& arr);
+
+   /// Get size
+   int size() const { return colors.size(); }
```
najlkin, line 61: Insisting to keep these small case? 🧐

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+private:
+   vector<RGBAf> colors;
+};
+
+/// Generates the texture data for a given palette, to be used in OpenGL
+class Texture
+{
+public:
+   /// The palette to create a texture of
+   Palette* const palette;
```
najlkin, line 87: `const Palette* const` or `const Palette &` would be better.

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   /// Repeat the palette multiple times (negative for reverse); cannot be 0
+   int Nrepeat_;
```
najlkin, line 89: No underscores please 🙂

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   /// Number of colors to discretize with (0 uses the original number of colors)
+   int Ncolors_;
+   /// Is texture smooth or discrete?
+   bool smooth;
+   /// Constructor - generates texture
```
najlkin, line 94: Lines between at least methods please 🙏

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   int size() const { return texture_data.size(); }
+   /// Get texture data
+   const vector<array<float,4>>& GetData() const { return texture_data; }
+   /// If true, all colors in palette are read in reverse
+   bool IsReversed() const { return Nrepeat_ < 0; }
+   /// Generates the texture data
+   void Generate();
+
+private:
+   int MAX_TEXTURE_SIZE;
```
najlkin, line 107: This should be renamed to match the naming convention and also could be static (inited to something invalid and then set by the first instance).

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   void AddPalette(Palette& palette);
+
+   /// Create a new palette with the given name and add it to the registry
+   void AddPalette(const string& name);
+
+   /// Returns true if name is unique
+   bool IsNameUnique(const string& name) const;
+
+   /// Get a palette pointer by index; if not found, returns last palette
+   Palette* get(int index) const;
```
najlkin, line 147: Also small case? 🙃

lib/palettes_base.hpp
```diff
@@ -0,0 +1,175 @@
...
+   /// Prints a summary (index + name) of all palettes
+   void PrintSummary(ostream& os = cout) const;
+
+   /// Prints all colors for all palettes
+   void PrintAll(ostream& os = cout) const;
+
+   /// Number of palettes in the registry
+   int NumPalettes() const { return palettes.size(); }
+
+   /* Loads palette(s) from a file. Format is:
```
najlkin, line 161:
```suggestion
   /// Loads palette(s) from a file.
   /** Format is:
```

lib/palettes.cpp
```diff
@@ -7703,16 +107,15 @@ void PaletteState::Init()
...
          std::lock_guard<std::mutex> lk{init_mtx};
-         Init_Palettes();
       }
 
-      GLuint paletteTexIds[Num_RGB_Palettes][2];
+      vector<array<GLuint, 2>> paletteTexIds(Palettes->NumPalettes());
       GLuint alphaTexId;
 
-      glGenTextures(Num_RGB_Palettes * 2, &(paletteTexIds[0][0]));
+      glGenTextures(Palettes->NumPalettes() * 2, &(paletteTexIds[0][0]));
```
najlkin, line 115: I thought this will go `Texture`, i.e., it will hold the actual OpenGL texture 🤔 . It would also save some memory, because `texture_data` are needed only temporarily to prepare the textures 😉 .

najlkin (commented):
Almost there 👍

lib/palettes_base.cpp
```diff
@@ -0,0 +1,337 @@
...
+      << setw(10) << g << " "
+      << setw(10) << b << " "
+      << setw(10) << a;
+}
+
+template <size_t N>
+Palette::Palette(const string& name,
+                 const array<array<float,3>,N>& arr) : name(name)
+{
+   colors.reserve(N);
```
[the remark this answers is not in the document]
justinlaughlin: thanks, changed

lib/palettes_base.cpp
```diff
@@ -0,0 +1,337 @@
...
+   return false;
+}
+
+Texture::Texture(Palette* palette, int Nrepeat_, int Ncolors_,
+                 bool smooth) : palette(palette), Nrepeat_(Nrepeat_),
+   Ncolors_(Ncolors_), smooth(smooth)
+{
+   // Get the maximum texture size
+   glGetIntegerv(GL_MAX_TEXTURE_SIZE, &MAX_TEXTURE_SIZE);
+   if (MAX_TEXTURE_SIZE < 4096)
```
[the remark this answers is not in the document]
najlkin: After a bit of archeology 🤠 , It was already in version 3.0 in `aux_vis.cpp` (3d6e0f71523f79ec6d09e0c64f856a99b4e3d68f) and in version 1.0 as 512 (6fdcb9580f997a2182dae1314bd52e7dfdc86611). The reason seems to be preallocation of some data arrays with this size, which we do not need anymore, so I would get rid of it 🙌

lib/palettes.cpp
```diff
@@ -7703,16 +107,15 @@ void PaletteState::Init()
       MaxTextureSize = std::min(MaxTextureSize, 4096);
       {
          std::lock_guard<std::mutex> lk{init_mtx};
-         Init_Palettes();
```
najlkin, line 7706: Remove the lock and mutex, it was here to protect the init, which we do not have anymore 🙌

lib/palettes_base.cpp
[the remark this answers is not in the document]
justinlaughlin: Nice find, thanks. I'll eliminate the magic number `4096` but keep `GL_MAX_TEXTURE_SIZE` since that is a GL requirement.

lib/palettes.cpp
justinlaughlin: done 👍

lib/palettes_base.hpp
justinlaughlin: done 👍

lib/palettes_base.hpp
justinlaughlin: changed

lib/palettes_base.hpp
justinlaughlin: I searched the repo and found both camelcase and snakecase for static non-const ints. I set it to snakecase and I also made this static.

lib/palettes_base.hpp
justinlaughlin: done 👍

lib/palettes.cpp
najlkin: mutex still remains 😬 , so it cannot compile...

lib/palettes_base.cpp
```diff
@@ -0,0 +1,336 @@
...
+{
+   for (const auto& color : colors)
+   {
+      if (color.a != 1.0) { return true; }
+   }
+   return false;
+}
+
+// Initialize
+int Texture::max_texture_size = 4096;
```
najlkin, line 91: The init below will not work with a non-zero value 🙃 . Besides, I would prefer something really invalid like `-1` to make it obvious.

justinlaughlin: oops; I meant to set that to 0 but was testing and forgot to revert.  I'll change to -1

najlkin: Better, but change the init condition accordingly 🙂

lib/palettes_base.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+// Initialize
+int Texture::max_texture_size = -1;
+
+Texture::Texture(const Palette* palette, int Nrepeat_, int Ncolors_,
+                 bool smooth) : palette(palette), Nrepeat_(Nrepeat_),
+   Ncolors_(Ncolors_), smooth(smooth)
+{
+   if (Texture::max_texture_size == 0)
```
najlkin, line 97:
```suggestion
   if (Texture::max_texture_size < 0)
```

lib/palettes.cpp
justinlaughlin: I also had this in mind but was holding off to wrap this up 😅.. but I agree with you that it would be cleaner. I'll make this change - some other things like global alpha texture (from `PaletteState::GenerateAlphaTexture`) will also need to be refactored similarly.

lib/palettes_base.cpp
justinlaughlin: thanks

lib/palettes_base.hpp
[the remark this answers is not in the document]
justinlaughlin: Done 👍

lib/palettes_base.cpp
justinlaughlin: Thats fair - we don't query by name yet but it will be nice in `pyglvis` and in glvis `evil` mode 😈 . I like the idea of overwriting because that seems like the more common scenario

lib/palettes.cpp
justinlaughlin: See [4d4be89](https://github.com/GLVis/glvis/pull/289/commits/4d4be8913d2cfd56fc65f60359cf5bcb1f6c01d2)

najlkin: Cool, step in good direction 👍 A few things:

- You forgot about `first_init`
- CI complains about ordering of inits in constructor
- Get rid of `PaletteTexIds`, textures can generate the ids by themselves. Not batched, but that does not matter much, I believe.
- The statics of `Texture` should be private. It uses only `rgba_internal` from the formats, alphas do not belong there. WIP, I guess. 🚧 
- Methods `SetCycle/Color` should be private as they set the texture to a semi-invalid state without update of the size. I do not see any separate usage, so maybe you may merge them with `UpdateTextureSize()`? 🤔

justinlaughlin: I think these are all addressed now, thanks!

lib/palettes_base.hpp
```diff
@@ -0,0 +1,227 @@
...
+   Texture(float matAlpha, float matAlphaCenter);
+
+   /// Initialize Static GL parameters
+   void InitStaticGL();
+
+   /// Get texture size.
+   int Size() { return tsize; }
+
+   /// Get the GL texture
+   GLuint Get() const { return texture; }
```
najlkin, line 113: The naming is a bit inconsistent here, either I would use `GetGLTexture()` or rename the method below to only `Generate()`.

lib/palettes.cpp
najlkin: Looks good here 😉

najlkin (commented):
Almost almost there 😄  Pls, do not forget about that loading above.

lib/palettes.cpp
```diff
@@ -7532,234 +46,63 @@ int PaletteState::ChoosePalette()
...
+   alpha_tex = Texture(1.0, 0.5);
 }
 
-static std::mutex init_mtx;
-
-void PaletteState::Init()
+void PaletteState::GenerateTextures()
 {
-   if (!first_init)
+   if ((int)(textures.size()) != Palettes->NumPalettes())
```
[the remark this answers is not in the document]
justinlaughlin: This is not used at the moment but you're right. I changed it to a more generic `reinitialize` argument that can be used in the future.

justinlaughlin:
Added some changes based on your feedback. Thanks!

lib/palettes.cpp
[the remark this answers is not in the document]
najlkin: That makes sense too :+1: . Maybe let us keep this open for future reference.

lib/palettes_base.cpp
najlkin: That overwriting looks simple here, just empty the existing one and the rest should work :wink:

lib/palettes_base.cpp
```diff
@@ -443,18 +443,16 @@ void PaletteRegistry::Load(const string& palette_filename)
...
-            cout << "Reading palette: (" << idx+1 << ") " << palname << endl;
-         }
-         else
-         {
-            cout << "Error reading palette: " << palname
-                 << ". Palette with same name already exists." << endl;
-            break;
+            cout << "Warning: palette name <" << palname
+                 << "> already exists. Overriding..." << endl;
+            palettes.erase(palettes.begin() + idx);
```
najlkin, line 450: As it is a good practice of premature optimization in MFEM, it would be more optimal if you do `palettes[idx] = as_unique<Palette>(palname)` 😄 . But for unification of the approach it is better like this. Would be better to wrap this call in a method, but also not super important.

najlkin (approved):

justinlaughlin:
Should this be squashed before merging? There is a lot of commits.

tzanio (approved):
Thanks a lot for this improvement, @justinlaughlin and @najlkin !

PR #296, opened 2024-07-23
CI misc improvements

justinlaughlin:
Changes:

 - Generate image diffs (abs and rel)
   - "abs" is the L2 of image diff in rgb space with a fixed scaling factor
   - "rel" is scaled by the max diff found in the images
   - PNGs are saved for "abs" and "rel"
   - An html is also generated using "rel". This lets you interactively zoom/pan to see where the test failed. It can only be viewed with an internet connection (this was done to save filesize)
 - Always run all tests (set `fail-fast: false`)
 - Rename artifacts to (hopefully) be less confusing
 - Misc refactoring
 - Separate unused code


tests/glvis_driver.py
```diff
@@ -72,31 +72,31 @@ def compare_images(baseline_file, output_file, expect_fail=False):
...
             print("[PASS] Images match.")
-    print("       actual ssim = {}, cutoff = {}".format(ssim, cutoff_ssim))
+    print(f"       actual ssim = {ssim}, cutoff = {cutoff_ssim}")
     return ssim >= cutoff_ssim if not expect_fail else ssim < cutoff_ssim
 
 # Function to test a given glvis command with a variety of key-based commands.
 # Not currently in use.
 def test_case(exec_path, exec_args, baseline, t_group, t_name, cmd):
-    print("Testing {0}:{1}...".format(t_group, t_name))
+    print(f"Testing {t_group}:{t_name}..."))
```
najlkin, line 81: Extra `)` here ❗

justinlaughlin: oops, I used a vim macro and it wasn't correct 😅. thanks

tzanio:
Should we add this to the checklist in https://github.com/GLVis/glvis/pull/284?

justinlaughlin:
I don't think its necessary since it is only features to help with development. I think the other PRs are very close / ready so I don't want to hold up the release. This could be merged in afterwards.

najlkin:
You may just rerun it, the Mac runners are extra crappy 😄 🤭

justinlaughlin:
Btw, has anyone seen this error before? I definitely saw it in a test last week (trying to find the job) but in the next test it was fixed and I'm not sure why.

```
6: *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'API misuse: setting the main menu on a non-main thread. Main menu contents should only be modified from the main thread.'
```

najlkin:
It happens randomly, I have seen that before. It is only on Mac 🍎🪱

justinlaughlin:
> You may just rerun it, the Mac runners are extra crappy 😄 🤭

oh, are you're telling me the mac tests are stochastic? 🤣

justinlaughlin:
needs a lot of cleanup but the basic idea for image diffs is there. (they are `html` files in the zip). The reason behind html is so the views can be interactive (via `plotly`)

e.g. `test.diff(zoomed).mfem-logo.html`
image

synchronized views
image


Right now the artifact size is huge because I'm generating it for everything, but in practice it'd only be generated when a test fails.

Still not sure why macos tests are failing 😞

justinlaughlin:
If anyone wants to try it out, controls are in top-right.

image

najlkin:
That is really cool 🤩 , but why is it so big? The normal PNGs have like hundreds of KB, while this has multiple MB. Also the UI is a little bit slow. Is that you store it uncompressed?

najlkin:
Regarding the failures, it seems this workaround does not work anymore for some reason or in just very stochastic way 😄 https://github.com/GLVis/glvis/blob/master/lib/sdl_main.cpp#L295-L309

najlkin:
Maybe it is not needed 🤔 , the commit d10bc970ef73e04be0747cedda32450292b16ac3 in  PR #174 says it is for version 2.0.12, but the runner uses SDL v2.30.5 😅 . Maybe try to comment it out and see what happens...

najlkin:
Omg, I maybe know what happens 😮 . It is something else totally. That function `GetMainThread()` is sometimes called from a non-main thread before it is in the main one, that is what it is complaining about, because constructor of `SdlMainThread` calls `SDL_Init()`, which fails then.

justinlaughlin:
> That is really cool 🤩 , but why is it so big? The normal PNGs have like hundreds of KB, while this has multiple MB. Also the UI is a little bit slow with such big files. Is that you store it uncompressed?

It is html for interactivity - I'm not sure how to reduce the file-size. We could just store the diff as a png as well, but I think it is helpful to directly compare all three images (baseline, output, diff). I'm not *too* concerned about size because in a normal use-case only one or a few of these should be generated. 

> Omg, I maybe know what happens 😮 . It is something else totally. That function GetMainThread() is sometimes called from a non-main thread before it is in the main one, that is what it is complaining about, because constructor of SdlMainThread calls SDL_Init(), which fails then.

That is a great find 😄

najlkin:
I mean HTML is fine, but can that HTML use PNG with lossless compression?

najlkin:
Yeeh, we are passing 🥳 Ok, I will make a PR for it :wink:.

najlkin:
How about `source` parameter of `plotly.graph_objects.Image`, so just storing PNGs and then generating the HTML for interactive plotting? If you include some switch for the files, there could just a single HTML, so no duplicity of the `plotly` JS source 😉 .

justinlaughlin:
> How about `source` parameter of `plotly.graph_objects.Image`, so just storing PNGs and then generating the HTML for interactive plotting? If you include some switch for the files, there could be just a single HTML, so no duplicity of the `plotly` JS source 😉 .

Sounds complicated 😅 if this was a core feature I would be more willing to optimize but this is just a matter of saving some MB on failed tests artifacts 🤷. Also the end-goal is to make the CI output easier to interpret/use not harder 😆.

najlkin:
I think having PNGs around is useful anyway and the easiest thing to interpret 🙂 . That HTML can be just a cherry on top 🍒  and if it is hard to add the switch then just generate a separate one for every image, that would not be a disaster 😉 .

justinlaughlin:
> I think having PNGs around is useful anyway and the easiest thing to interpret 🙂 . That HTML can be just a cherry on top 🍒 and if it is hard to add the switch then just generate a separate one for every image, that would not be a disaster 😉 .

yes agreed :). All the PNGs will still be there, this `*diff.html` will only show up when a test fails so it is easier to troubleshoot why.

najlkin:
with the diff loaded from a PNG as well, right? 😛

justinlaughlin:
> How about `source` parameter of `plotly.graph_objects.Image`, so just storing PNGs and then generating the HTML for interactive plotting? If you include some switch for the files, there could be just a single HTML, so no duplicity of the `plotly` JS source 😉 .

oops my bad I don't know why when I first read this I interpreted it differently, this seems pretty straightforward. will check if it reduces size

justinlaughlin:
> How about `source` parameter of `plotly.graph_objects.Image`, so just storing PNGs and then generating the HTML for interactive plotting? If you include some switch for the files, there could be just a single HTML, so no duplicity of the `plotly` JS source 😉 .

Thanks for the idea! Works great :) filesize is much smaller now (still MBs but not tens of MBs)

najlkin:
So is the diff PNG as well? Can we see it somewhere? I am curious 🤤
Besides that, there seems to be like a top level `.` directory in the tar, could you remove that? 🤔

najlkin:
Some filename error, it could not find files to upload 😕

najlkin:
There is no HTML 😢

justinlaughlin:
> There is no HTML 😢

I had it generate for all cases earlier for testing, but now I set it only to generate if there is a test failure. Filesize is still ~5MB per `html`. Do you think we should just include it?

najlkin:
Just for testing, or maybe just attach an example file here so we can see it 😉

najlkin:
Still no HTML, I was not that bad idea to test it 😄

justinlaughlin:
> Just for testing, or maybe just attach an example file here so we can see it 😉

I changed the `glvis/data` hash to make an artifact, I'll switch it back later. There is an html file under `mesh-explorer`

najlkin:
Ahaa, like that, but why is it so big? It seems it still includes the picture data. I unzippzed only the HTML and it shows the pictures without the PNGs 😅

justinlaughlin:
> Ahaa, like that, but why is it so big? It seems it still includes the picture data. I unzippzed only the HTML and it shows the pictures without the PNGs 😅

It is the bundled source code. There is an option under `figure.write_html()` called `include_plotlyjs='cdn'`. That reduces the filesize down to around the size of the images - but then the `html` can only be used online. We could do that?

najlkin:
No, I meant to use `file://test.nominal....png` in `source`, so no data in HTML. It is not possible?

justinlaughlin:
> No, I meant to use `file://test.nominal....png` in `source`, so no data in HTML. It is not possible?

Can we discuss over slack real quick? might be easier

najlkin:
Maybe we could add here the that enabling of the style checks for PR CIs. Or separate? 🤔

justinlaughlin:
> Maybe we could add here the that enabling of the style checks for PR CIs. Or separate? 🤔

If you want to add it, up to you. Otherwise imo, separate PR. I need to work on another project so if we want to include this in 4.3 I say ship it as-is.

tzanio:
I just merged `master` in here, and everything looks good.

Are we ready to merge this?

najlkin (approved):
LGTM, good work 👍

v-dobrev (approved):

tzanio (approved):
Please mention in `CHANGELOG`

PR #301, opened 2024-07-29
Fixed dynamic loading of parallel quadratures

najlkin:
This PR fixes dynamic loading of parallel quadratures, which did not work with the "parallel ... quadrature" keywords after the first stream, but expected a special "pquadrature" keyword. Moreover, the loader of parallel grid functions was not checking the second keyword, so without any warning the parallel quadratures were treated as grid functions, leading to crashes 💣 . With this fix, "parallel ... quadrature" can be used consistently and the second keyword is checked properly.

Reported by: @vladotomov

### DIFF
--- a/lib/threads.cpp
+++ b/lib/threads.cpp
@@ -914,6 +914,10 @@ void communication_thread::execute()
                   qf_array.SetSize(nproc);
                   qf_array[proc] = new QuadratureFunction(mesh_array[proc], isock);
                }
+               else
+               {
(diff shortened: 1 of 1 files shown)


najlkin:
Cannot dowload `gperf`, because gnu.org is down 🦬 🎯 .

vladotomov (approved):
Now it works, thank you @najlkin.

tzanio:
I restarted the CI and it all passes now 👍

tzanio (approved):

PR #302, opened 2024-07-31
Fixed use-after-delete with dynamic streams and switching quadratures

najlkin:
This PR fixes access to the old mesh in `VisualizationScene::NewMeshAndSolution()` shortly after it was deleted in `StreamState::SetNewMeshAndSolution()` by the move of the new data there and similarly when switching quadrature function representations.

❓  This bug was introduced in #286 .

### DIFF
--- a/lib/stream_reader.cpp
+++ b/lib/stream_reader.cpp
@@ -327,7 +327,7 @@ void StreamState::SetQuadSolution(QuadSolution type)
 void StreamState::SwitchQuadSolution(QuadSolution type, VisualizationScene *vs)
 {
    unique_ptr<Mesh> old_mesh;
-   //we must backup the refined mesh to prevent its deleting
+   // we must backup the refined mesh to prevent its deleting
    if (mesh_quad.get())
    {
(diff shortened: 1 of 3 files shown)


lib/stream_reader.cpp
```diff
@@ -324,6 +324,20 @@ void StreamState::SetQuadSolution(QuadSolution type)
    quad_sol = type;
 }
 
+void StreamState::SwitchQuadSolution(QuadSolution type, VisualizationScene *vs)
+{
+   unique_ptr<Mesh> old_mesh;
+   //we must backup the refined mesh to prevent its deleting
```
tzanio, line 330:
```suggestion
   // we must backup the refined mesh to prevent its deleting
```

tzanio (approved):

najlkin:
Testing it should be easy, just print out the pointers of the old and new mesh in `VisualizationScene::NewMeshAndSolution()` and in `StreamState` to see what goes where and what is deleted 😉 .

tzanio:
Can we get another review please?

vladotomov (approved):

PR #303, opened 2024-08-06
Added a test for LOR quad visualization.

najlkin:


najlkin:
@tzanio , do you know how to add a new baseline? I added some dummy pictures, but I do not see any related output in the artifacts 🤔

PR #305, opened 2024-08-07
Fix the CMake 'install' target on macOS to sign bundled dependencies.

pazner:
Also, in the 'release' GitHub action, use 'runner.arch' instead of 'amd64'.


v-dobrev (approved):

tzanio (approved):
Thank you for the quick fix @v-dobrev and @pazner !

Issue #308, opened 2024-08-14
Exclamation mark does not work with some keyboard layouts


najlkin:
Exclamation mark for changing the texture type does not work with some keyboard layouts (like Czech).

Here is the output from tracking of key presses with Czech layout:
```
Event: SDL_KEYDOWN sym=1073742053 mod=4098
Event: SDL_KEYDOWN sym=167 mod=4098
Event: SDL_TEXTINPUT text[0..3]=33 0 0 0 (as codes 0-255)
Event: SDL_KEYUP sym=167 mod=4098
Event: SDL_KEYUP sym=1073742053 mod=4096
```
and for reference English layout (which works):
```
Event: SDL_KEYDOWN sym=1073742049 mod=4097
Event: SDL_KEYDOWN sym=49 mod=4097
Event: SDL_TEXTINPUT text[0..3]=33 0 0 0 (as codes 0-255)
Event: SDL_KEYUP sym=49 mod=4097
Event: SDL_KEYUP sym=1073742049 mod=4096
```
❓  This probably comes from #294

najlkin:
This is wrong https://github.com/GLVis/glvis/blob/2b8a8ec1315969da2c54e453c4e4fd3c33b1a85e/lib/sdl.cpp#L276 but I do not know how to fix it. The thing is the key (167) is out of the range, but with Shift/Caps Lock it is translated to the range (33).

najlkin:
Translation by `SDL_GetScancodeName()` would work for me, as it at least return the key with English layout, which is apostrophe, but it is not universal 🤔

najlkin:
Actually, it should be fine, because US keyboard matches ASCII table 😉

najlkin:
This would be the fix working for me:
```c++
const char *scan_name = SDL_GetScancodeName(ks.scancode);
if ((scan_name[0] >= 32 && scan_name[0] < 127) && scan_name[1] == '\0'
    && (ks.mod & (KMOD_CTRL | KMOD_ALT | KMOD_GUI)) == 0)
```

PR #309, opened 2024-08-14
Fixed Shift combinations with different keyboard layouts

najlkin:
This is a possible fix for #308 . Please test it!

### DIFF
--- a/lib/sdl.cpp
+++ b/lib/sdl.cpp
@@ -627,7 +627,7 @@ void SdlWindow::signalKeyDown(SDL_Keycode k, SDL_Keymod m)
    event.type = SDL_KEYDOWN;
    event.key.windowID = window_id;
    event.key.keysym.sym = k;
-   event.key.keysym.scancode = SDL_GetScancodeFromKey(k);
+   event.key.keysym.scancode = SDL_GetScancodeFromKey(tolower(k));
    event.key.keysym.mod = m;
    queueEvents({ event });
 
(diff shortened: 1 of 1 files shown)


justinlaughlin:
Something funky going on with test 11 `mesh-explorer` on mac 👀 

image

najlkin:
What is that !? 🙃

najlkin:
Can you test it @justinlaughlin ? I do not have a Mac 🍏 ...

justinlaughlin:
> Can you test it @justinlaughlin ? I do not have a Mac 🍏 ...

I don't have one either unfortunately 🏠🐧

najlkin:
Ok, I maybe know what happens, `SdlWindow::signalKeyDown()` generates also these key down events, but does not initialize the scan code, which did not matter before, but matters now 🤔 .

najlkin:
Ok, still some keys are handled in `keyDownEvent()` instead of `textInputEvent()`, but it does not matter, I would say, as the called functions are identified by the key code, which is always correct from `signalKeyDown()`.

tzanio:
I can confirm that the latest commit shows the mesh-explorer test correctly on my mac

tzanio (approved):
I tested some more, and this appears to work as before with the standard layout on Mac

justinlaughlin:
FWIW, tested `mesh-explorer` on WSL and it seems to be fine.

justinlaughlin (approved):

najlkin:
@justinlaughlin  Do you have different layouts than the US?

justinlaughlin:
> @justinlaughlin Do you have different layouts than the US?

Just USA 🦅

PR #310, opened 2024-08-15
Fixes for glvis-js

justinlaughlin:
Some of the changes in `v4.3` broke portions of `aux_js.cpp` because they aren't using the updated constructors, enums, etc. This PR makes updates so we can build `glvis-js` again.

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -92,8 +92,8 @@ StreamState::FieldType display(const StreamState::FieldType field_type,
    vs = nullptr;
 
    double mesh_range = -1.0;
-   if (field_type == StreamState::FieldType::SCALAR ||
-       field_type == StreamState::FieldType::MESH)
+   if (field_type == StreamState::FieldType::SCALAR
+       || field_type == StreamState::FieldType::MESH)
    {
       if (stream_state.grid_f)
       {
@@ -105,11 +105,12 @@ StreamState::FieldType display(const StreamState::FieldType field_type,
          if (stream_state.normals.Size() > 0)
          {
             vs = vss = new VisualizationSceneSolution(*stream_state.mesh, stream_state.sol,
-                                                      &stream_state.normals);
+                                                      stream_state.mesh_quad.get(), &stream_state.normals);
          }
          else
          {
-            vs = vss = new VisualizationSceneSolution(*stream_state.mesh, stream_state.sol);
+            vs = vss = new VisualizationSceneSolution(*stream_state.mesh, stream_state.sol,
+                                                      stream_state.mesh_quad.get());
          }
          if (stream_state.grid_f)
          {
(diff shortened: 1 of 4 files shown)


lib/aux_js.cpp
```diff
@@ -207,11 +208,11 @@ int display(const int field_type, std::stringstream & commands, const int w,
       }
       if (stream_state.mesh->SpaceDimension() == 2 && field_type == 2)
       {
-         SetVisualizationScene(vs, 2);
+         SetVisualizationScene(vs, 2, stream_state.keys.c_str());
       }
       else
       {
-         SetVisualizationScene(vs, 3);
+         SetVisualizationScene(vs, 3, stream_state.keys.c_str());
```
najlkin, line 215: This looks unrelated to the changes, is it some bugfix?

justinlaughlin: Nope - was copying stuff over from `glvis.cpp` but this shouldn't be here. Thanks :)

najlkin:
Btw, how is this compiled? And another one, how about adding support for quadrature functions? 🤔

justinlaughlin:
> Btw, how is this compiled? And another one, how about adding support for quadrature functions? 🤔

Basically, follow the build section in the README: https://github.com/GLVis/glvis-js

Although, in order to build I needed to make a change in mfem ([PR](https://github.com/mfem/mfem/pull/4453)) and also use an older emscripten version. I want to setup CI similar to mfem->pymfem so it will rebuild glvis-js on pushes to glvis master 😅

tzanio (approved):

justinlaughlin:
@najlkin Is it okay to always pass `StreamState.mesh_quad` e.g. like in 
```c++
vs = vss = new VisualizationSceneSolution(*stream_state.mesh, stream_state.sol,
                                          stream_state.mesh_quad.get(), &stream_state.normals);
```

even if we aren't using quadrature visualization? If so then I can remove the new constructors.

najlkin:
Yes, this is how it works in `glvis.cpp` now. Stream state does not set the coarse mesh unless it is that quadrature function OR it is 1D, which was the bug in the other PR 😅 .

najlkin:
Btw, a funny thing is that I do not use that coarse mesh at all 😄 , it is only for indication that you want visualize the coarse mesh lines or not, but it is conceptually cleaner to do it this way, because it is just an implementation detail that there is no dependency, but you could use it somehow in principle.

justinlaughlin:
Okay, sounds good to me!

najlkin:
Besides that, it seems to me that GLVis live should be able to load and visualize quadratures with this update automatically, since we are loading stream files directly 😮 .

najlkin (approved):

justinlaughlin:
Thanks for reviewing 😎

PR #312, opened 2024-08-19
Add PLASIMO colormap and make it the default

Rickbude:


Rickbude:
PR opened by mistake, my bad..

PR #313, opened 2024-08-28
Js enum binding

justinlaughlin:
We made some changes in https://github.com/GLVis/glvis/pull/310 to use `StreamState::FieldType` but the `enum` also needed to be bound in emscripten.

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -516,7 +516,6 @@ em::val getPNGByteArray()
 EMSCRIPTEN_BINDINGS(js_funcs)
 {
    em::enum_<StreamState::FieldType>("FieldType")
-   .value("ONE", StreamState::FieldType::ONE)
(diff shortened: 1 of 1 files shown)


najlkin (approved):
Not like I could check anything much, but looks sensible :wink:.

tzanio (approved):

justinlaughlin:
> Not like I could check anything much, but looks sensible 😉.

Our lovely new [CI](https://github.com/GLVis/glvis-js/actions/runs/10606600123) will build `glvis-js` for you 😄. All you need to do is use the artifact and run `examples/basic.html`.

tzanio:
> > Not like I could check anything much, but looks sensible 😉.
> 
> Our lovely new [CI](https://github.com/GLVis/glvis-js/actions/runs/10606600123) will build `glvis-js` for you 😄. All you need to do is use the artifact and run `examples/basic.html`.

That's awesome work and really great improvement, @justinlaughlin !

PR #314, opened 2024-08-30
Quadrature visualization support for glvis-js

justinlaughlin:
Eventually, we should refactor `aux_js.cpp` a bit so that there isn't so much repeated code with `glvis.cpp`. For now, this is just following the same design pattern to see if we can get quadrature support in `glvis-js`.

### DIFF
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -338,13 +338,6 @@ int updateStream(std::string stream)
    std::string data_type;
    ss >> data_type;
 
-   if (data_type != "solution")
-   {
-      std::cerr << "unsupported data type '" << data_type << "' for stream update" <<
-                std::endl;
-      return 1;
-   }
-
    StreamState new_state;
    new_state.ReadStream(ss, data_type);
 
--- a/lib/aux_js.cpp
+++ b/lib/aux_js.cpp
@@ -44,6 +44,9 @@ namespace js
 
 using namespace mfem;
 
+// switch representation of the quadrature function
+void SwitchQuadSolution();
+
 //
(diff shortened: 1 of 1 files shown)


justinlaughlin:
I think this fixed it? To test

```bash
git clone git@github.com:GLVis/glvis-js.git
cd glvis-js
git checkout ci-build-10626571587
open examples/basic.html # or open with a browser
# open glvis-data/streams/quadrature-lor.saved and press Q
```

tzanio:
I can confirm `Q` works now. Thanks @justinlaughlin ~

tzanio (approved):

najlkin (commented):
Yes, works for me too :wink: 🚀  I agree it would be better to unify the code with `glvis.cpp`, but there are some subtleties different, so it is not that easy.
However, I noticed that `updateStream()` has a hardcoded check for `solution`, so that needs to be fixed for mesh and quadrature 🤔

justinlaughlin:
> Yes, works for me too 😉 🚀 I agree it would be better to unify the code with `glvis.cpp`, but there are some subtleties different, so it is not that easy. However, I noticed that `updateStream()` has a hardcoded check for `solution`, so that needs to be fixed for mesh and quadrature 🤔

`updateStream()` uses `StreamState::SetNewMeshAndSolution` which it looks like you updated already 🙂. I went ahead and removed the check; I don't see any downstream calls specific to `data_type`. I tested the build `ci-build-10706880188` and it seems okay to me.

najlkin (approved):
LGTM, the check in `updateStream` was superfluous, I think :wink:.

Issue #326, opened 2025-03-02
crash displaying vector fields with vdim != mesh dims and vdim > 1


helloworld922:
Reproducer:

```c++
#include <mfem.hpp>

int main(int argc, char **argv) {
  using namespace mfem;
  int order = 2;
  // glvis will segfault for vdims != mesh_dims and vdims > 1
  int vdims = 3;
  int mesh_dims = 2;
  Mesh mesh;
  switch (mesh_dims) {
  case 1:
    mesh = Mesh::MakeCartesian1D(1);
    break;
  case 2:
    mesh =
        Mesh::MakeCartesian2D(1, 1, Element::Type::QUADRILATERAL, false, 1, 1);
    break;
  case 3:
    mesh = Mesh::MakeCartesian3D(1, 1, 1, Element::Type::HEXAHEDRON, 1, 1, 1);
    break;
  }

  L2_FECollection l2_coll(order, mesh_dims);
  FiniteElementSpace vec_space(&mesh, &l2_coll, vdims);
  Array<int> offset(2);
  offset[0] = 0;
  offset[1] = offset[0] + vec_space.GetVSize();
  BlockVector data(offset, Device::GetMemoryType());
  GridFunction vec_gf;
  vec_gf.MakeRef(&vec_space, data, offset[0]);

  socketstream sout;
  char vishost[] = "localhost";
  int visport = 19916;
  sout.open(vishost, visport);
  if (!sout) {
    std::cout << "Unable to connect to GLVis server at " << vishost << ':'
              << visport << std::endl;
    std::cout << "GLVis visualization disabled.\n";
    return -1;
  }
  sout.precision(8);
  sout << "solution\n" << mesh << vec_gf;
  sout << std::flush;
}
```
This causes glvis to crash (segfault?)

najlkin:
Checking the code, it is not supported now as the space dimension is used to determine what visualization scene object is constructed. However, maybe we could treat it as 2D surface and use the 3D scene object for it 🤔 .

najlkin:
Not that easy, it causes crashes at many places in the 3D scene object as it assumes it has space dimension 3. It would be a lot of work to go through all methods and check for compatibility with 2D meshes.
Another option is to extrude the mesh to 3D, but there might be attributes, boundaries and so on, so it is not that easy as with 1D meshes, which are extruded to 2D for visualization.
So I do not know, I would suggest just splitting to 2D vdim=2 and 2D scalar components for now.

[...]

Issue #291, opened 2024-07-17
ND_R2D visualization with glvis fails


Rickbude:
This is a companion issue to the identically-named issue in MFEM: [https://github.com/mfem/mfem/issues/4403](https://github.com/mfem/mfem/issues/4403)

Loading a solution that is expressed in terms of ND_R1D, ND_R2D, RT_R1D and RT_R2D elements gives an error that looks like:

```
Verification failed: (dim >= 1 && dim <= 3) is false:
 --> ND_FECollection requires 1 <= dim <= 3.
 ... in function: mfem::ND_FECollection::ND_FECollection(int, int, int, int)
 ... in file: /home/rick/tmp/mfem/fem/fe_coll.cpp:2708

Aborted (core dumped)
```
 
This error is caused by missing code in MFEM (see the linked issue for more information). Adding the missing code resolves things on the MFEM side, but now glvis throws a segmentation fault. This is the trace obtained with gdb:

```
#0  0x00007ffff5e195aa in _XEventsQueued () from /lib64/libX11.so.6
#1  0x00007ffff5df5e87 in XCheckIfEvent () from /lib64/libX11.so.6
#2  0x00007ffff6a0c3a5 in ?? () from /lib64/libSDL2-2.0.so.0
#3  0x00007ffff6951442 in ?? () from /lib64/libSDL2-2.0.so.0
#4  0x00007ffff6951867 in ?? () from /lib64/libSDL2-2.0.so.0
#5  0x000000000044a9ca in SdlMainThread::DispatchSDLEvents (this=this@entry=0x933e00 <GetMainThread()::inst>) at /home/rick/tmp/glvis/lib/sdl_main.cpp:191
#6  0x000000000044aea3 in SdlMainThread::MainLoop (this=0x933e00 <GetMainThread()::inst>, server=<optimized out>) at /home/rick/tmp/glvis/lib/sdl_main.cpp:140
#7  0x0000000000442ab4 in SdlWindow::StartSDL (server_mode=<optimized out>) at /home/rick/tmp/glvis/lib/sdl.cpp:85
#8  0x000000000042fb99 in SDLMainLoop (server_mode=<optimized out>) at /home/rick/tmp/glvis/lib/aux_vis.cpp:55
#9  0x000000000041f3d2 in main (argc=<optimized out>, argv=<optimized out>) at /home/rick/tmp/glvis/glvis.cpp:1400
```

Rickbude:
[...]

PR #327, opened 2025-03-04
Visualization of 2D grid functions with 3 vector dimensions

najlkin:
This addresses the issue of elements with vdim=3 in 2D (or 1D) from #326 and vector elements RT/ND_R2D spaces from #291 .
◀️ dependency: RT/ND_R2D support requires GLVis compiled with mfem/mfem#4731 .
image

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -25,6 +25,10 @@ Version 4.3.3 (development)
 
 - Fixed a bug where discrete textures would not repeat.
 
+- Added support for 2D data (mesh and grid function) using scalar finite
+  elements with 3 vector dimensions or vector finite elements with 3 range
+  dimensions (requires MFEM v4.8).
+
 
 Version 4.3.2 released on Sep 27, 2024
 ======================================
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -223,7 +223,8 @@ bool GLVisInitVis(StreamState::FieldType field_type,
    }
    else if (field_type == StreamState::FieldType::VECTOR)
    {
-      if (stream_state.mesh->SpaceDimension() == 2)
+      if (stream_state.mesh->SpaceDimension() == 2 && (!stream_state.grid_f ||
+                                                       stream_state.grid_f->VectorDim() == 2))
       {
          if (stream_state.grid_f)
          {
@@ -235,11 +236,38 @@ bool GLVisInitVis(StreamState::FieldType field_type,
                                               stream_state.solv, stream_state.mesh_quad.get());
          }
       }
-      else if (stream_state.mesh->SpaceDimension() == 3)
+      else if (stream_state.mesh->SpaceDimension() == 3 || (stream_state.grid_f &&
+                                                            stream_state.grid_f->VectorDim() == 3))
       {
          if (stream_state.grid_f)
          {
             stream_state.ProjectVectorFEGridFunction();
+            // Extrude 2D3V grid functions to 3D
+            if (stream_state.mesh->SpaceDimension() < 3)
+            {
+               Mesh *mesh3d = new Mesh(*stream_state.mesh);
+               const FiniteElementSpace *fesx = stream_state.mesh->GetNodalFESpace();
+               const FiniteElementCollection *fecx = (fesx)?(fesx->FEColl()):(NULL);
+               if (fecx)
+               {
+                  mesh3d->SetCurvature(fecx->GetOrder(),
+                                       fecx->GetContType() == FiniteElementCollection::DISCONTINUOUS,
+                                       3);
+               }
+               else
+               {
+                  mesh3d->SetCurvature(1, false, 3);
+               }
+               FiniteElementSpace *fes2d = stream_state.grid_f->FESpace();
+               FiniteElementSpace *fes3d = new FiniteElementSpace(*fes2d, mesh3d);
+               GridFunction *gf3d = new GridFunction(fes3d);
+               *gf3d = *stream_state.grid_f;
+               stream_state.grid_f->MakeOwner(NULL);
+               gf3d->MakeOwner(const_cast<FiniteElementCollection*>(fes2d->FEColl()));
+               stream_state.SetGridFunction(gf3d);
+               delete fes2d;
+               stream_state.SetMesh(mesh3d);
+            }
             vs = new VisualizationSceneVector3d(*stream_state.grid_f,
                                                 stream_state.mesh_quad.get());
          }
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -223,8 +223,7 @@ bool GLVisInitVis(StreamState::FieldType field_type,
    }
    else if (field_type == StreamState::FieldType::VECTOR)
    {
-      if (stream_state.mesh->SpaceDimension() == 2 && (!stream_state.grid_f ||
-                                                       stream_state.grid_f->VectorDim() == 2))
+      if (stream_state.mesh->SpaceDimension() == 2)
       {
          if (stream_state.grid_f)
          {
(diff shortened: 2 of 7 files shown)


najlkin:
@helloworld922 , can you test this? It worked for me with the reproducer 😉 .

helloworld922:
It doesn't crash now, but now it doesn't update the vector data on subsequent plots for time series data.

Here's an updated reproducer:

```
#include <mfem.hpp>

int main(int argc, char **argv) {
  using namespace mfem;
  int order = 2;
  int vdims = 3;
  int mesh_dims = 2;
  Mesh mesh;
  int nelems1d = 10;
  switch (mesh_dims) {
  case 1:
    mesh = Mesh::MakeCartesian1D(nelems1d);
    break;
  case 2:
    mesh =
        Mesh::MakeCartesian2D(nelems1d, nelems1d, Element::Type::QUADRILATERAL, false, 1, 1);
    break;
  case 3:
    mesh = Mesh::MakeCartesian3D(nelems1d, nelems1d, nelems1d, Element::Type::HEXAHEDRON, 1, 1, 1);
    break;
  }

  L2_FECollection l2_coll(order, mesh_dims);
  FiniteElementSpace vec_space(&mesh, &l2_coll, vdims);
  Array<int> offset(2);
  offset[0] = 0;
  offset[1] = offset[0] + vec_space.GetVSize();
  BlockVector data(offset, Device::GetMemoryType());
  GridFunction vec_gf;
  vec_gf.MakeRef(&vec_space, data, offset[0]);

  mfem::VectorFunctionCoefficient vec_coeff(
      vec_space.GetVDim(),
      std::function<void(const mfem::Vector &, mfem::real_t, mfem::Vector &)>(
          [&](const mfem::Vector &pos, mfem::real_t time, mfem::Vector &res) {
            res[0] = sin(pos[0] + time);
            res[1] = cos(pos[1] + time);
            res[2] = time;
          }));
  mfem::real_t time = 0;
  mfem::real_t dt = 1e-2;
  vec_coeff.SetTime(time);
  vec_gf.ProjectCoefficient(vec_coeff);
  vec_gf.SyncAliasMemory(data);

  socketstream sout;
  char vishost[] = "localhost";
  int visport = 19916;
  sout.open(vishost, visport);
  if (!sout) {
    std::cout << "Unable to connect to GLVis server at " << vishost << ':'
              << visport << std::endl;
    std::cout << "GLVis visualization disabled.\n";
    return -1;
  }
  sout.precision(8);
  sout << "solution\n" << mesh << vec_gf;
  sout << "window_title 't = " << time << "'\n";
  sout << "pause\n";
  sout << std::flush;

  long long step = 0;
  while (true) {
    ++step;
    time = step * dt;
    vec_coeff.SetTime(time);
    vec_gf.ProjectCoefficient(vec_coeff);
    vec_gf.SyncAliasMemory(data);
    if (!sout) {
      break;
    }
    sout << "solution\n" << mesh << vec_gf;
    sout << "window_title 't = " << time << "'\n" << std::flush;
  }
}
```

najlkin:
Ok, it should work now 👍 I moved things around a bit.

helloworld922:
It looks like it works for L2 spaces now. Thanks!

najlkin:
@mlstowell , any error message? Maybe try it with MFEM checked out at https://github.com/mfem/mfem/pull/4731 . It is required for that RT/ND_R1/2D elements and should not be necessary without them, but still better to try 😉 .

mlstowell:
Nothing useful. The reproducer shows
```
Error in send(): Broken pipe
Error in send(): Broken pipe
```

GLVis shows:
```
OpenGL Visualization
Testing if OpenGL core profile window can be created...success!
Handle for window created.
High-dpi display detected: using window scaling: 2 x 2
2025-03-05 11:49:05.706 glvis[95478:5520998] +[IMKClient subclass]: chose IMKClient_Modern
Using GLEW 2.2.0
Using GL 4.1 Metal - 89.3
Loading CoreGLDevice...
Using GLSL 410
Window should be up
Font pattern 'Ubuntu Light:style=Regular' matched no fonts.
Font pattern 'Ubuntu:style=Regular:weight=80' matched no fonts.
Font pattern 'OpenSans:style=Regular' matched no fonts.
Font pattern 'DejaVu Sans:style=Book:width=Normal' matched no fonts.
Font pattern 'DejaVu LGC Sans:style=Book:width=Normal' matched no fonts.
Font pattern 'Bitstream Vera Sans:style=Roman' matched no fonts.
Font pattern 'FreeSans:style=Medium' matched no fonts.
Font pattern 'Ubuntu Mono:style=Regular' matched no fonts.
Font pattern 'DejaVu Sans Mono:style=Book' matched no fonts.
Font pattern 'DejaVu LGC Sans Mono:style=Book' matched no fonts.
Font pattern 'Helvetica:style=Regular' matched fonts:
 - Helvetica:style=Regular,標準體,Almindelig,Κανονικά,Normaali,Courant,רגיל,Szabályos,Regolare,レギュラー,일반체,Regulier,Normal,Обычный,ปกติ,Reguler,Звичайний,Thường,रेग्युलर,Biasa,常规体,عادي:weight=80:file=/System/Library/Fonts/Helvetica.ttc:index=0:scalable=True
Screen DPI: w = 255 ppi, h = 255 ppi (255 x 255)
Fonts use fixed ppi: 192 x 192
Loaded font: /System/Library/Fonts/Helvetica.ttc, index: 0, height: 12
GlVisFont::LoadFont(): font texture dimensions are (1622, 32)
Initializing textures.
VisualizationSceneVector::VisualizationSceneVector
Trace/BPT trap
```

mlstowell:
Oops, I see the problem...

mlstowell:
It's working for me now. I'll run some more testing after lunch but I do like this approach. Thanks @najlkin!

najlkin:
So have you tracked that down? The thing is GLVis cannot require the latest MFEM. It is ok for new features like that RT/ND_R1/2D elements, but not the core things  🤔 .

mlstowell:
@najlkin, have a look at PR #329 which improves the vector scaling. I'm still seeing other issues but this may correct some of them

mlstowell:
Another issue I'm seeing is that sometimes the GLVis window is empty. I believe this sometimes happens when the z-coordinates are uninitialized and the scaling of the image is not computed correctly. I was looking at setting the z-coordinate in `StreamState::Extrude2D3VMeshAndSolution` but I'm not sure if that's the best solution.

najlkin:
> Another issue I'm seeing is that sometimes the GLVis window is empty. I believe this sometimes happens when the z-coordinates are uninitialized and the scaling of the image is not computed correctly. I was looking at setting the z-coordinate in `StreamState::Extrude2D3VMeshAndSolution` but I'm not sure if that's the best solution.

I have not seen it, but then it is a bug of `Mesh::SetCurvature()`.

najlkin:
Hmm, `XYZ_VectorFunction` in `mesh.cpp`, which is used for the projection, looks fine 🤔 .

mlstowell:
> > Another issue I'm seeing is that sometimes the GLVis window is empty. I believe this sometimes happens when the z-coordinates are uninitialized and the scaling of the image is not computed correctly. I was looking at setting the z-coordinate in `StreamState::Extrude2D3VMeshAndSolution` but I'm not sure if that's the best solution.
> 
> I have not seen it, but then it is a bug of `Mesh::SetCurvature()`.

Yes, exactly. I was debating where it should be fixed. I think you're correct. If `Mesh::SetCurvature` is being used to increase the space dimension then it must ensure that the new coordinate is valid everywhere.

najlkin:
We may include it to mfem/mfem#4731 , but I do not know where it is coming from. Maybe something relies on the vertices, which are still 2D, while `Nodes` are 3D 🤔 However, I do not know what and where. Valgrind does not show anything useful.

najlkin:
Yeah, it seems that `SetVerticesFromNodes(Nodes);` at the end of `SetCurvature()` solves that, but I am not sure it is the right thing to do. It would be better to know what uses that invalid vertices 🤔 .

najlkin:
Ok, I added such fix to mfem/mfem#4731 , because many methods in GLVis rely on vertices instead of nodes and there is no way to know what dimension they have after the change as the higher ones are simply uninitialized 💣 .

PR #329, opened 2025-03-06
2d3v-vis suggestion 

mlstowell:
This produces better scaling for some vector display types (cycled using 'v') on 2D meshes embedded in 3D space.

### DIFF
--- a/lib/vsvector3d.hpp
+++ b/lib/vsvector3d.hpp
@@ -22,6 +22,7 @@ protected:
 
    Vector *solx, *soly, *solz;
    int drawvector, scal_func;
+   double volume;
    gl3::GlDrawable vector_buf;
    gl3::GlDrawable displine_buf;
 
--- a/lib/vsvector3d.hpp
+++ b/lib/vsvector3d.hpp
@@ -22,7 +22,7 @@ protected:
 
    Vector *solx, *soly, *solz;
    int drawvector, scal_func;
-   double volume;
+   double mesh_volume;
    gl3::GlDrawable vector_buf;
    gl3::GlDrawable displine_buf;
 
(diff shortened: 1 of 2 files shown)


najlkin:
Cool, I noticed that vector arrows do not appear. This solves it nicely 👍 :
image
However, I am a bit worried if it should be applied generally:

- It breaks backward compatibility
- Slim objects in 2/3D had arrows proportional to the bounding box, so the arrows were always visible, but with this they are proportional to the volume, so might be tiny, but better matching the size of the body of the object. It is a choice what we want 🤔 .

mlstowell:
We could change this so that the maximum of the bounding box volume or the length/area/volume of the mesh is used. This would fix backward compatibility for 3D meshes or lower dimensional meshes which are not aligned with the coordinate axes. But it would also avoid the issues arising from 3D bounding boxes with zero volume.

najlkin:
I pushed to `2d3v-vis` a simpler solution, just calculating bounding area instead of volume for embedded 2D objects. That keeps backward compatiblity and vector arrows proportional to the bounding box.

mlstowell:
That works as well.

najlkin:
Yeah, both work fine 😉 , so I do not know. 3D volume will be always smaller than that of the bounding box, unless there is some edge case of curved volumes 🤔 .

lib/vsvector3d.hpp
```diff
@@ -22,6 +22,7 @@ class VisualizationSceneVector3d : public VisualizationSceneSolution3d
 
    Vector *solx, *soly, *solz;
    int drawvector, scal_func;
+   double volume;
```
najlkin, line 25:
```suggestion
   double mesh_volume;
```

najlkin (approved):
Ok, I think this is a bit cleaner in the sense the reader will not be left wondering why 2D area is considered 3D volume 😄 . After the CI finishes, I would merge it.

PR #335, opened 2025-05-06
Refactoring and scripts commands help

najlkin:
Refactored the code base, separating scripts handling and introducing a help for script commands (which can be placed on the website too). At the moment it is:

> Available commands are:
        mesh <file> - Visualize the mesh.
        solution <mesh> <solution> - Visualize the solution.
        psolution <np> <mesh prefix> <keep attributes> <solution prefix> - Visualize the distributed solution.
        quadrature <mesh> <quadrature> - Visualize the quadrature.
        pquadrature <np> <mesh prefix> <keep attributes> <quadrature prefix> - Visualize the distributed quadrature.
        data_coll_mesh <type> <data coll> - Visualize the mesh from data collection.
        data_coll_field <type> <data coll> <field> - Visualize the field from data collection.
        data_coll_quad <type> <data coll> <quad> - Visualize the Q-field from data collection.
        data_coll_cycle <cycle> - Preset the cycle of the data collection.
        data_coll_protocol <protocol> - Preset the protocol of the data collection.
        screenshot <file> - Take a screenshot, saving it to the file.
        viewcenter <x> <y> - Change the viewcenter.
        perspective <on/off> - Turn on or off perspective projection.
        light <on/off> - Turn on or off light.
        view <theta> <phi> - Change the solid angle of view.
        zoom <zoom> - Change the zoom factor.
        shading <flat/smooth/cool> - Change the shading algorithm.
        subdivisions <times> <dummy> - Change the refinement level.
        valuerange <min> <max> - Change the value range.
        autoscale <off/on/value/mesh> - Change the autoscale algorithm.
        levellines <min> <max> <num> - Set the level lines.
        axis_numberformat '\<format\>' - Set the axis number format.
        colorbar_numberformat '\<format\>' - Set the colorbar number format.
        window <x> <y> <w> <h> - Set the position and size of the window.
        keys <key> - Send the control key sequence.
        palette <index> - Set the palette index.
        palette_repeat <times> - Set the repetition of the palette.
        toggle_attributes <1/0> [[<1/0>] ...]; - Toggle visibility of the attributes.
        rotmat <[0,0]> <[1,0]> ... <[3,3]> - Set the rotation matrix.
        camera <cam[0]> ... <cam[2]> <dir[0]> ... <dir[2]> <up[0]> ... <up[2]> - Set the camera position, direction and upward vector.
        scale <scale> - Set the scaling factor.
        translate <x> <y> <z> - Set the translation coordinates.
[...]

Issue #315, opened 2024-09-04
Stream commands unavailable in glvis-js


justinlaughlin:
Most stream commands are not available in glvis-js/pyglvis because `aux_js.cpp/display` is handled differently.

See: https://github.com/GLVis/pyglvis/issues/49

> Unfortunately, `valuerange` is the only other command that works - `glvis-js` has different machinery for handling the stream (`glvis/lib/aux_js.cpp/display`). In particular, see these [lines](https://github.com/GLVis/glvis/blob/master/lib/aux_js.cpp#L61-L77). As an example, you can add `zoom 50` at the end of a `.saved` file and open it with `glvis` vs `glvis-js`.

PR #336, opened 2025-05-12
[WIP] EGL headless visualization

najlkin:
This PR adds headless visualization from scripts using EGL standard.

↪️ dependencies:

- #335


najlkin:
I have tried headless CI tests with EGL and some pass smoothly, but other fail at initialization. There seems to be some issue on the runners like limited video memory or something on the runners, so it is very random. You may compare [this run](https://github.com/GLVis/glvis/actions/runs/15031953362/job/42306719434) and [this](https://github.com/GLVis/glvis/actions/runs/15031953362/job/42246214799). I have removed the tests and only building with EGL support is done on Ubuntu.

najlkin:
Ok, the headless server works nicely now. It should be tested in combination with parallel applications on clusters. If IP/TCP sockets cause some problems, we could try to use Unix sockets locally on the node 🤔 .

tzanio:
> Ok, the headless server works nicely now. It should be tested in combination with parallel applications on clusters. If IP/TCP sockets cause some problems, we could try to use Unix sockets locally on the node 🤔 .

That's awesome, thanks @najlkin !

tzanio:
@justinlaughlin, what happens if we load a steam with "screenshot" command in the web version?

najlkin:
I think an error 😁 , because glvis-js supports only `keys` and `valuerange` commands apart from the data loading ones (see #315 ). But you may enter `keys S` to get a screenshot. I wonder what will happen then 🤔 .

najlkin:
Hmm, at least in `basic.html` nothing happens after `S`. It would be too nice to get a file 🙃 . I guess it violates some security policies and the code does not get a writing file handle 🚫 .

najlkin:
But in glvis live, it works😮 . However, there is a special code for it, handling specifically `S`, calling `saveScreenshot()` in `index.js`. You get `glvis.png` file for download. It shares `SaveAsPNG()` with the main code, so there is a potential for headless operation if you are thinking that way 😉 .

CHANGELOG
```diff
@@ -14,6 +14,16 @@ Version 4.4.1 (development)
 
 Unlike previous GLVis releases, this version requires a C++17 compiler.
 
+- Added headless (no GUI) visualization, relying on the EGL standard. It is
```
tzanio: Mention CGL

tzanio (approved):
Very excited about this 👍

CHANGELOG
najlkin: Done 😉

CHANGELOG
```diff
@@ -14,6 +14,16 @@ Version 4.4.1 (development)
 
 Unlike previous GLVis releases, this version requires a C++17 compiler.
 
+- Added headless (no GUI) visualization, relying on the EGL (Linux) or CGL
+  (MacOS) interface. It is available through '-hl' command-line option or
+  'headless' script command (before the visualization commands). End of the
+  loaded stream or script file end the visualization in this mode. Note GLVis
```
mlstowell: This doesn't sound quite right. Perhaps `The end of the loaded stream or script file ends the visualization in this mode.` would be better.

najlkin: Done, thanks.

lib/threads.hpp
```diff
@@ -58,7 +56,8 @@ class GLVisCommand
       PALETTE_REPEAT = 20,
       LEVELLINES = 21,
       AXIS_NUMBERFORMAT = 22,
-      COLORBAR_NUMBERFORMAT = 23
+      COLORBAR_NUMBERFORMAT = 23,
```
justinlaughlin: Will conflict with new commands in #344 - https://github.com/GLVis/glvis/pull/344/files#diff-6ffdc23ac638868988ae692f69ffd3b5962166fdf531b9a3c81db85ccc9e6389R59-R61. I guess we'll see what merges first

justinlaughlin (approved):

najlkin:
Resolved 🫡 🚀

tzanio:
@GLVis/developers -- please speak up if you have any concerns about this PR which also includes significant refactoring and improvements.

If there are no concerns, we plan to merge this tomorrow (Jan 30th).

najlkin:
The problem with glvis-js has been resolved 👍

tzanio (approved):
Great stuff, @najlkin !

PR #338, opened 2025-05-28
CGL headless visualization

najlkin:
This is an addition to #336 with CGL (Core OpenGL) backend instead of EGL for MacOS 🍎 .

✔️ testing:
- [x] CI build
- [x] manual build
- [x] headless functionality


najlkin:
@camierjs , it seems I have accidentally uncovered something here by the error check in `initGLEW()`. Would you have time to check where it comes from?

najlkin:
A question to everyone, do we want to make compilation with CGL default on Mac? 🤔  There is no additional dependency, so there is no need to keep it optional (unlike EGL, which is usually packed with OpenGL on Linux, but there is a slight chance it is not).

tzanio:
Thanks @najlkin!

Lets review this 👍

INSTALL
```diff
@@ -117,6 +117,8 @@ Some important variables for CMake are:
 
 - GLVIS_USE_EGL: Use EGL for headless rendering. Default is "OFF".
 
+- GLVIS_USE_CGL: Use CGL for headless rendering on macos. Default is "OFF".
```
camierjs: `ON` on `macos`?

camierjs (approved):

INSTALL
najlkin: Updated 😉

Issue #106, opened 2020-07-23
Displays element and vertex numberings, but not dof numberings


rw-anderson:
It should probably display dof numberings as well.  Maybe something like this for Q2. :)

![dofs](https://user-images.githubusercontent.com/11507994/88258789-cbc2c280-cc75-11ea-99c0-45d7a0fd4a5f.png)

tzanio:
I agree this could be useful :).

It could also be nice to be able to display the DOF numbering for the GridFunction that is being visualized.

rw-anderson:
Yes, I'm grabbing the DOFs from the GridFunction.  Here's an L2 GridFunction.  Still some work to complete this.  I'll push a branch.

![star-L2](https://user-images.githubusercontent.com/11507994/88338709-9954ac00-cced-11ea-8219-80f6c3de5cce.png)

tzanio:
ping: @rw-anderson

tzanio:
@rw-anderson, do you want to add this as part of the upcoming `glvis-4.1` or should we postpone it to the next release?

hughcars:
Just wanted to say this would be extremely helpful for debugging feature additions in mfem. Even more useful if it would be able to handle NC dofs.

camierjs:
This issue should be resolved by the changes in PR #339, which adds support for displaying DOF numberings.

PR #339, opened 2025-05-28
DOFs numberings displays

camierjs:
This PR adds DOFs numbering.

The `n` key now cycles in 2D: `NONE` → `ELEMENTS` → `EDGES` → `VERTICES` → `DOFS`.

Should solve #106 

GLVis_s01

Parallel numbering is now by default 'local' to each rank and can be toggled back to 'global' with `Alt-n`:
output
(`mpirun -n 3 ./ex6p -m ../data/star-mixed.mesh -o 2 -md 200`)

### DIFF
--- a/lib/data_state.cpp
+++ b/lib/data_state.cpp
@@ -645,13 +645,13 @@ void DataState::ComputeDofsOffsets(std::vector<mfem::GridFunction*> &gf_array)
 
    DenseMatrix pointmat;
    Array<int> dofs, vertices;
-   for (int i = 0, g_e = 0; i < nprocs; i++)
+   for (int rank = 0, g_e = 0; rank < nprocs; rank++)
    {
-      const GridFunction *gf = gf_array[i];
+      const GridFunction *gf = gf_array[rank];
       const FiniteElementSpace *l_fes = gf->FESpace();
       Mesh *l_mesh = l_fes->GetMesh();
       // store the dofs numbers as they are fespace dependent
-      auto &offset = (*offsets)[i];
+      auto &offset = (*offsets)[rank];
       for (int l_e = 0; l_e < l_mesh->GetNE(); l_e++, g_e++)
       {
 #ifdef GLVIS_DEBUG
@@ -664,7 +664,7 @@ void DataState::ComputeDofsOffsets(std::vector<mfem::GridFunction*> &gf_array)
             xs += pointmat(0,j), ys += pointmat(1,j);
          }
          xs /= nv, ys /= nv;
-         offset.exy_map[ {l_e, i} ] = {xs, ys};
+         offset.exy_map[ {g_e, rank} ] = {xs, ys};
 #endif // end GLVIS_DEBUG
          l_fes->GetElementDofs(l_e, dofs);
          l_fes->AdjustVDofs(dofs);
@@ -673,8 +673,8 @@ void DataState::ComputeDofsOffsets(std::vector<mfem::GridFunction*> &gf_array)
             offset[ {g_e, k} ] = dofs[k];
          }
       }
-      if (i + 1 == nprocs) { continue; }
-      auto &next = (*offsets)[i+1];
+      if (rank + 1 == nprocs) { continue; }
+      auto &next = (*offsets)[rank+1];
       // for NE, NV and NEdges, we accumulate the values
       next.nelems = offset.nelems + l_mesh->GetNE();
       next.nedges = offset.nedges + l_mesh->GetNEdges();
--- a/lib/vssolution.cpp
+++ b/lib/vssolution.cpp
@@ -1860,7 +1860,7 @@ void VisualizationSceneSolution::PrepareElementNumbering()
          if (!el_attr_to_show[mesh->GetAttribute(e)-1]) { continue; }
          center_ir.IntPoint(0) =
             Geometries.GetCenter(mesh->GetElementBaseGeometry(e));
-         GetRefinedValues (e, center_ir, values, pointmat);
+         GetRefinedValues(e, center_ir, values, pointmat);
          const double xc = pointmat(0,0), yc = pointmat(1,0), uc = values(0);
          const double dx = 0.05 * GetElementLengthScale(e);
          const double xx[3] = {xc, yc, uc};
@@ -1872,8 +1872,7 @@ void VisualizationSceneSolution::PrepareElementNumbering()
             const int rank = mesh->GetAttribute(e) - 1;
             MFEM_VERIFY(rank >= 0 && rank < (int)offsets->size(),
                         "Invalid rank");
-            const int l_e = offset(e);
-            const auto &xy = (*offsets)[rank].exy_map.at({l_e,rank});
+            const auto &xy = (*offsets)[rank].exy_map.at({e,rank});
             MFEM_VERIFY(fabs(xy.x - xc) < eps && fabs(xy.y - yc) < eps,
                         "Element center does not match expected position");
          }
@@ -1888,8 +1887,8 @@ void VisualizationSceneSolution::PrepareElementNumbering()
       for (int e = 0; e < mesh->GetNE(); e++)
       {
          if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
-         mesh->GetPointMatrix (e, pointmat);
-         mesh->GetElementVertices (e, vertices);
+         mesh->GetPointMatrix(e, pointmat);
+         mesh->GetElementVertices(e, vertices);
          const int nv = vertices.Size();
          ShrinkPoints(pointmat, e, 0, 0);
          double xs = 0.0, ys = 0.0, us = 0.0;
@@ -1909,7 +1908,7 @@ void VisualizationSceneSolution::PrepareElementNumbering()
             const int rank = mesh->GetAttribute(e) - 1;
             MFEM_VERIFY(rank >= 0 && rank < (int)offsets->size(),
                         "Invalid rank");
-            const auto &xy = (*offsets)[rank].exy_map.at({offset(e), rank});
+            const auto &xy = (*offsets)[rank].exy_map.at({e, rank});
             MFEM_VERIFY(fabs(xy.x - xs) < eps && fabs(xy.y - ys) < eps,
                         "Element center does not match expected position");
          }


tzanio:
To clarify: "DOF" here means the numbering of the Mesh nodes (as a high-order grid function), right?

lib/data_state.hpp
```diff
@@ -13,9 +13,15 @@
...
-#include <vector>
 #include <memory>
-#include "mfem.hpp"
+
+#include <mfem.hpp>
+using mfem::Mesh;
+using mfem::Vector;
+using mfem::GridFunction;
+using mfem::QuadratureFunction;
+using mfem::DataCollection;
```
najlkin, line 23: Why to do this? It would be fine in a cpp file, but a header file it is "contagious", spreading in an uncontrolled way 🫤 . Maybe scope it by putting it to `DataState`? 😉

najlkin:
Js seems to be working: https://github.com/GLVis/glvis-js/actions/runs/15313952476 😉 Thanks for the cleanup 👍 I need to take a deeper look...

lib/data_state.hpp
camierjs: Trying to avoid the 'using namespace' in the header and not poluting with too many 'mfem::'. 
Good idea moving it inside the scope.

najlkin:
I tested that and it works nicely with 2D scalars on conforming meshes, but a few things to do:

- [ ] update README
- [ ] update Changelog
- [ ] the current option could be written to console
- [ ] 2D vector grid functions do something strange after pressing `N`, should be fixed
- [ ] example 6 crashes glvis after switching to the second option, should be fixed

najlkin:
@tzanio , it shows DOFs (in the sense of finite element nodes) for the visualized grid function. In the case of mesh visualization, the nodes of the coloring grid function.

lib/data_state.hpp
camierjs: Scoping was not working, I cleaned up the header files by using `mfem::` specifically.

lib/gl/renderer_core.hpp
```diff
@@ -11,11 +11,17 @@
...
+
+#include <unordered_map>
+
 #include "renderer.hpp"
 #include "shader.hpp"
 
 namespace gl3
 {
+
+using namespace resource;
```
najlkin: Maybe when you have removed `using` from `renderer.hpp`, remove it also here (and put it to the cpp file maybe).

lib/gl/types.cpp
```diff
@@ -20,8 +19,8 @@ void GlDrawable::addCone(float x, float y, float z,
                          float vx, float vy, float vz,
                          float cone_scale)
 {
-   double rhos  = sqrt (vx*vx+vy*vy+vz*vz);
-   float phi   = acos(vz/rhos);
+   double rhos  = sqrt(vx*vx+vy*vy+vz*vz);
+   float phi   = acos (vz/rhos);
    float theta = atan2 (vy, vx);
```
najlkin:
```suggestion
   double rhos  = sqrt(vx*vx+vy*vy+vz*vz);
   float phi    = acos(vz/rhos);
   float theta  = atan2(vy, vx);
```

CHANGELOG
```diff
@@ -9,6 +9,12 @@
                              https://glvis.org
 
 
+Version 4.5 (development)
+=========================
+
+- Implemented DOF numbering, pressing the `n` key in 2D mode cycles through:
```
najlkin:
```suggestion
- Implemented DOF numbering, pressing the `n` key in 2D mode cycles through:
```
```suggestion
- Implemented DOF numbering, pressing the 'n' or 'N' key in 2D mode cycles through:
```

lib/aux_vis.cpp
```diff
@@ -435,7 +434,7 @@ void MyExpose()
 }
 
 
-thread_local Array<void (*)()> IdleFuncs;
+thread_local mfem::Array<void (*)()> IdleFuncs;
```
najlkin, line 440: If this is the only use of MFEM, I would suggest to change it to `std::vector` and remove the include.

lib/data_state.hpp
```diff
@@ -84,7 +86,7 @@ struct DataState
 
    /// Set a mesh (plain pointer version)
    /** Note that ownership is passed from the caller.
-       @see SetMesh(std::unique_ptr<mfem::Mesh> &&pmesh) */
+       @see SetMesh(std::unique_ptr<Mesh> &&pmesh) */
```
najlkin: Not completely sure, but I think doxygen follows the scopes of the code, so it might not find the symbols. Perhaps revert these as well.

lib/material.cpp
```diff
@@ -9,62 +9,64 @@
 // terms of the BSD-3 license. We welcome feedback and contributions, see file
 // CONTRIBUTING.md for details.
 
-#include "material.hpp"
 #include <array>
```
najlkin: Already in the header file.

lib/palettes_base.cpp
```diff
@@ -12,15 +12,16 @@
 #include "palettes_base.hpp"
 #include "palettes_default.cpp"
```
najlkin:
```suggestion
```
This is awful 😬 . Just remove it here please and add it to the makefile among source files, that is all what is needed (I have tested that 😉 ).

lib/vssolution.cpp
```diff
@@ -586,14 +577,23 @@ void VisualizationSceneSolution::ToggleDrawElems()
...
    Mesh *old_m = mesh;
    mesh = new_m;
    mesh_coarse = new_mc;
-   sol = new_sol;
-   rsol = new_u;
+   MFEM_VERIFY(new_sol->Size() == mesh->GetNV(),
+               "New solution vector size does not match the mesh node count.");
+   delete sol;
+   sol  = new Vector(mesh->GetNV());
+   SetGridFunction(*new_u);
```
najlkin: This is not right in the way it was intended. The scene can be constructed from scalar data without a grid function and this method was supposed to update either the scalar data or the grid function. So if `new_u` is null, the scalar data should be just copied. However, is it really necessary? The `DataState` is supposed to hold the data, so just a pointer was passed before. Secondly, `SetGridFunction()` should not be called without a grid function (will give a segfault, I guess 💣 ).

lib/vssolution.cpp
```diff
@@ -1998,39 +1997,134 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
-         const double m[2] = {0.5 * (p(0,0) + p(0,1)), 0.5 * (p(1,0) + p(1,1))};
-         // TODO: figure out something better...
-         double u = LogVal(0.5 * ((*sol)(vertices[0]) + (*sol)(vertices[1])));
+   const int ne = mesh->GetNE();
+   auto *rsol_fes = rsol->FESpace();
+   const auto *rsol_fec = rsol_fes->FEColl();
+   FiniteElementSpace rdof_fes(mesh, rsol_fec);
+   FiniteElementSpace flat_fes(mesh, rsol_fec->Clone(1), rsol_fes->GetVDim());
+   // filter out unsupported basis types
+   const bool force_non_comforming = sol->Size() != flat_fes.GetNDofs();
```
najlkin: Is this reliable? 🤔

lib/vssolution.cpp
najlkin: Ohh, I also noticed this has broken vector fields visualization. With them, `sol` is not nodal values, but the result of vec to scalar function.

najlkin: If you are concerned about ownership of `sol`, look at refactoring in #335 . I have given the visualization scene access to `DataState` (among other things, because it is just an illusion it depends only on the things in the current constructor). So you may safely use the whole content of that structure.

lib/aux_vis.cpp
camierjs: We are already pulling `data_state.hpp` from `threads.hpp` and there are some calls (`Union`, `DeleteFirst`) that might not translate directly with `std::vector`.

lib/vssolution.cpp
camierjs: The DOFs positions are reliable, however the values height won't follow the 'flat' values, as in some other cases right now. It's only for the ND & RT basis types.

najlkin: In other words, the condition could be if the finite elements are scalar and the construction of the finite element space could be skipped right away.

lib/aux_vis.cpp
najlkin: Ok, let us leave it.

lib/vssolution.cpp
camierjs: That will certainly help the overall ownerships  !

makefile
```diff
@@ -254,9 +254,10 @@ ALL_SOURCE_FILES = \
  lib/gl/shader.cpp lib/gl/types.cpp lib/aux_js.cpp lib/aux_vis.cpp      \
  lib/coll_reader.cpp lib/data_state.cpp lib/file_reader.cpp             \
  lib/font.cpp lib/gl2ps.c lib/gltf.cpp lib/material.cpp                 \
- lib/openglvis.cpp lib/palettes.cpp lib/palettes_base.cpp lib/sdl.cpp   \
- lib/sdl_helper.cpp lib/sdl_main.cpp lib/sdl_windows.cpp                \
- lib/sdl_x11.cpp lib/stream_reader.cpp lib/threads.cpp lib/vsdata.cpp   \
+ lib/openglvis.cpp lib/palettes.cpp lib/palettes_base.cpp 				\
+ lib/palettes_default.cpp lib/sdl.cpp lib/sdl_helper.cpp 				\
+ lib/sdl_main.cpp lib/sdl_windows.cpp lib/sdl_x11.cpp 					\
+ lib/stream_reader.cpp lib/threads.cpp lib/vsdata.cpp   				\
```
najlkin: Spaces please 🙂

lib/vssolution.cpp
camierjs: I pushed a proper way to handle numbering for vector fe and flat/smooth shading.

lib/vssolution.cpp
najlkin: But why did you change the vector to an own copy instead of a pointer? If we assume there is always a vector available at your dispose (which is not very obvious now, but with `DataState` it is guaranteed), could not we change it back to the external data model?

najlkin: Could you try to revert this and merge #343 ?

camierjs: `done` in [3f80e0f](https://github.com/GLVis/glvis/pull/339/commits/3f80e0ff0e5d4f0b9a4bd149e8a294449347c4a6)

lib/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ list(APPEND SOURCES
   material.cpp
   openglvis.cpp
   palettes.cpp
+  palettes_default.cpp
```
najlkin, line 26: It is already below 👇

lib/palettes_base.cpp
```diff
@@ -13,14 +13,16 @@
 #include "gl/renderer.hpp"
 
 using namespace std;
+#include <iomanip>
+#include <fstream>
```
najlkin, line 16: Move before `using`?

lib/material.cpp
najlkin: Why not to remove it?

glvis.cpp
```diff
@@ -171,6 +174,7 @@ bool GLVisInitVis(StreamCollection input_streams)
             // the 'jet-like' palette is used in 2D, see vssolution.cpp).
             vs->palette.SetIndex(4);
          }
+         vs->SetDataOffsets(stream_state.offsets.get());
```
najlkin, line 178: I do not like these calls after constructor, if we are merging this one first, we must leave it here temporarily, but #335 introduces `Window` structure with the `DataState`, so this can be eliminated.

lib/data_state.hpp
```diff
@@ -93,6 +120,9 @@ struct DataState
        @see SetMesh(std::unique_ptr<mfem::Mesh> &&pmesh) */
    void SetMesh(mfem::Mesh *mesh);
 
+   /// Compute the dofs offsets from the grid function vector
+   void ComputeDofsOffsets(std::vector<mfem::GridFunction*> &gf_array);
+
```
najlkin: Maybe move it down a bit to/after the grid function methods.

lib/data_state.cpp
```diff
@@ -57,36 +58,82 @@ DataState &DataState::operator=(DataState &&ss)
...
+         double xs = 0.0, ys = 0.0;
+         for (int j = 0; j < nv; j++)
+         {
+            xs += pointmat(0,j), ys += pointmat(1,j);
+         }
+         xs /= nv, ys /= nv;
+         offset.exy_map[ {l_e, i} ] = {xs, ys};
+#endif // end GLVIS_DEBUG
+         l_fes->GetElementDofs(l_e, dofs);
+         l_fes->AdjustVDofs(dofs);
```
najlkin: DOFs do not have orientation, so this is unnecessary.

lib/data_state.cpp
```diff
@@ -57,36 +58,82 @@ DataState &DataState::operator=(DataState &&ss)
...
+         // Store elements centers
+         l_mesh->GetPointMatrix(l_e, pointmat);
+         const int nv = pointmat.Width();
+         double xs = 0.0, ys = 0.0;
+         for (int j = 0; j < nv; j++)
+         {
+            xs += pointmat(0,j), ys += pointmat(1,j);
+         }
+         xs /= nv, ys /= nv;
+         offset.exy_map[ {l_e, i} ] = {xs, ys};
```
najlkin: It should be checked here we are in 2D actually 🤔 .

lib/data_state.cpp
```diff
@@ -57,36 +58,82 @@ DataState &DataState::operator=(DataState &&ss)
...
+#endif // end GLVIS_DEBUG
+         l_fes->GetElementDofs(l_e, dofs);
+         l_fes->AdjustVDofs(dofs);
+         for (int k = 0; k < dofs.Size(); k++)
+         {
+            offset[ {g_e, k} ] = dofs[k];
+         }
+      }
+      if (i + 1 == nprocs) { continue; }
+      auto &next = offsets->operator[](i+1);
```
najlkin:
```suggestion
      auto &next = (*offsets)[i+1];
```

lib/vssolution.cpp
```diff
@@ -1825,169 +1833,167 @@ double VisualizationSceneSolution::GetElementLengthScale(int k)
...
-   else
+   auto offset = [&](const int e)
    {
-      PrepareElementNumbering1();
-   }
-}
+      if (!offsets) { return e; }
+      if (legacy_parallel_numbering)
+      {
+         // std::cout << "\x1b[31m[legacy_parallel_numbering]\x1b[m";
```
najlkin: Maybe remove these...

lib/vssolution.cpp
```diff
@@ -1998,39 +2004,128 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+   else if (shading == Shading::Noncomforming)
+   {
+      for (int e = 0; e < mesh->GetNE(); e++)
+      {
+         if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
+         mesh->GetElementEdges(e, edges, edges_ori);
+         const auto dx = 0.05 * GetElementLengthScale(e);
+         const auto geom = mesh->GetElementBaseGeometry(e);
+         MFEM_VERIFY(geom == Geometry::TRIANGLE || geom == Geometry::SQUARE,
+                     "Only TRIANGLE and SQUARE geometries are supported.");
```
najlkin: Rather show an error in console and return from this routine than killing GLVis altogether.

lib/vssolution.cpp
```diff
@@ -1998,39 +2004,128 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
 
-      for (int i = 0; i < edges.Size(); i++)
-      {
-         mesh->GetEdgeVertices(edges[i], vertices);
+   d_nums_buf.clear();
 
-         p.SetSize(mesh->Dimension(), vertices.Size());
-         p.SetCol(0, mesh->GetVertex(vertices[0]));
-         p.SetCol(1, mesh->GetVertex(vertices[1]));
+   auto *rsol_fes = rsol->FESpace();
```
najlkin: `rsol` may not exist in which case nodal scalar data are visualized, so this should be linear H1 collection/space. However, it will coincide with the vertex numbering, so it may default to that routine alternatively.

lib/vssolution.cpp
```diff
@@ -1998,39 +2004,128 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+   FiniteElementSpace rdof_fes(mesh, rsol_fes->FEColl());
+   H1_FECollection h1_fec(1, mesh->Dimension());
+   FiniteElementSpace h1_fes(mesh, &h1_fec);
+   MFEM_VERIFY(sol->Size() == h1_fes.GetNDofs(),
+               "Flat space does not match the solution size");
+   GridFunction h1_sol(&h1_fes, sol->GetData());
```
najlkin: This is not needed for non-conforming shading.

najlkin: Connected to this, do not call `GetRefinedValues()` for flat/smooth shading, because the values are not used anyway, but also the transformation should not be used as it is misplacing the DOFs applying curvilinear mapping on them, while these shadings use straight edges and affine/bilinear mapping only. Besides, it expects `rsol` to exist.

lib/CMakeLists.txt
camierjs: removed in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/palettes_base.cpp
camierjs: done in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

glvis.cpp
camierjs: Added a comment to follow up after #335 is merged

lib/data_state.hpp
camierjs: Moved after the Grid/Quad functions in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/data_state.cpp
camierjs: done in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/vssolution.cpp
camierjs: Removed in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/vssolution.cpp
camierjs: Error output & `return` in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/vssolution.cpp
camierjs: I split the Flat/Smooth vs. Noncomforming to avoid the extra call and use H1/rsol fespaces in respective cases in [2e1a271](https://github.com/GLVis/glvis/pull/339/commits/2e1a271cb2844b33ecdf7c10c58fb899ec253a63)

lib/threads.cpp
```diff
@@ -931,6 +937,7 @@ void communication_thread::execute()
             if (gf_array.size() > 0)
             {
                tmp.SetGridFunction(new GridFunction(tmp.mesh.get(), gf_array.data(), nproc));
+               if (!keep_attr) { tmp.ComputeDofsOffsets(gf_array); }
```
najlkin: Looking at this construction, how about making `ComputeDofsOffsets()` internal and call it from a special version of `SetGridFunction()` for arrays of grid functions? 👍  There are such versions for quadratures and complex functions, so it would be consistent and guarantees the data state has offsets for parallel solutions.

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,141 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+                  "Flat space does not match the solution size");
+      GridFunction h1_sol(&h1_fes, sol->GetData());
+
+      for (int e = 0; e < mesh->GetNE(); e++)
+      {
+         if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
+         const auto dx = 0.05 * GetElementLengthScale(e);
+         const auto &ir = h1_fes.GetFE(e)->GetNodes();
+         h1_fes.GetElementDofs(e, dofs);
+         h1_fes.AdjustVDofs(dofs);
```
najlkin: Also unnecessary for DOFs.

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,141 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+         for (int q = 0; q < ir.GetNPoints(); q++)
+         {
+            const real_t z = h1_sol.GetValue(e, ir.IntPoint(q));
+            const real_t x[3] = {tr(0,q), tr(1,q), z};
```
najlkin, line 2120: `tr` is uninitialized! 💣

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,141 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+      FiniteElementSpace h1_fes(mesh, &h1_fec);
+      MFEM_VERIFY(sol && sol->Size() == h1_fes.GetNDofs(),
+                  "Flat space does not match the solution size");
+      GridFunction h1_sol(&h1_fes, sol->GetData());
+
+      for (int e = 0; e < mesh->GetNE(); e++)
+      {
+         if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
+         const auto dx = 0.05 * GetElementLengthScale(e);
+         const auto &ir = h1_fes.GetFE(e)->GetNodes();
```
najlkin: This will not show DOFs of a high-order grid function when using smooth/flat shading. It should take the nodes from `rsol` if it exists and this should be backup if it does not.

lib/threads.cpp
najlkin: That is a good precaution, because it is easy to miss, like you did in `FileReader`, for example 😅 . Also we may centralize the logic for setting the element/boundary attributes 👍 .

lib/data_state.cpp
camierjs: Checked in https://github.com/GLVis/glvis/pull/339/commits/cc724b1092cb50f710ec9c0e66f08bc42a4c7913

lib/threads.cpp
camierjs: done in https://github.com/GLVis/glvis/pull/339/commits/f8558325f840cf95774445b538cdc02a89decd3f

lib/vssolution.cpp
camierjs: Split the cases, added the Flat or Smooth w/ rsol in https://github.com/GLVis/glvis/pull/339/commits/37326ad3b7bc6b769a3ef293553eae5e00d03bd2

lib/vssolution.cpp
camierjs: done in https://github.com/GLVis/glvis/pull/339/commits/37326ad3b7bc6b769a3ef293553eae5e00d03bd2

lib/threads.cpp
```diff
@@ -930,7 +936,7 @@ void communication_thread::execute()
             tmp.SetMesh(new Mesh(mesh_array.data(), nproc));
             if (gf_array.size() > 0)
             {
-               tmp.SetGridFunction(new GridFunction(tmp.mesh.get(), gf_array.data(), nproc));
+               tmp.SetGridFunction(gf_array, nproc);
```
najlkin, line 940: This relies on the `keep_attr` attribute of the new state, which might not be consistent with the current (in `keep_attr` variable). So line `tmp.keep_attr = keep_attr;` is needed here, or actually a little bit higher would be better (it is rather mesh related, not gf) 😉 . (It is a kind of bug that was here for long time, but it did not matter as `keep_attr` was not used anywhere after the initial construction)

lib/file_reader.cpp
najlkin, line 1: Do not forget to use the array variant of `SetGridFunction()` here on line 195 😅 .

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,175 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
+         GridFunction h1_sol(&h1_fes, sol->GetData());
 
-         double xx[3] = {m[0], m[1], u};
-         DrawNumberedMarker(f_nums_buf, xx, xs, edges[i]);
+         for (int e = 0; e < mesh->GetNE(); e++)
+         {
+            if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
+            const auto dx = 0.05 * GetElementLengthScale(e);
+            const auto &ir = rsol_fes->GetFE(e)->GetNodes();
+            GetRefinedValues(e, ir, vals, tr, true);
```
najlkin: Oh no, I did not mean it like this. Sorry for confusion if there was some 😅 `GetRefinedValues()` should not be used at all for these shadings, because the values are not used, but also the coordinates of the nodes in the physical space should not be, because it applies a high-order curvilinear mapping on them, while the element is drawn just with an affine/bilinear mapping only. The nodes in the *reference* space should be taken from the `rsol` FES, like it is here on the line above. That means the cases with/without `rsol` can be merged (but up to you), the only difference should be `rsol_fes->GetFE(e)->GetNodes()` vs. `h1_fes.GetFE(e)->GetNodes()`.

najlkin: btw, you may see the difference on: `./glvis -saved tests/data/streams/quad.saved`

lib/threads.cpp
camierjs: done in [a6c80cf](https://github.com/GLVis/glvis/pull/339/commits/a6c80cfd6833fbf60997b2aa0c7b406d509bfe2d)

lib/file_reader.cpp
camierjs: done in [e3c907f](https://github.com/GLVis/glvis/pull/339/commits/e3c907fd1e10e1d79db594673856c29bcf0652ff)

lib/vssolution.cpp
camierjs: simplified in [9b919cf](https://github.com/GLVis/glvis/pull/339/commits/9b919cf6cf3ac3c6bff7b6cf6cb3f2fff5d1fa71)

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,145 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
 
-         const double m[2] = {0.5 * (p(0,0) + p(0,1)), 0.5 * (p(1,0) + p(1,1))};
-         // TODO: figure out something better...
-         double u = LogVal(0.5 * ((*sol)(vertices[0]) + (*sol)(vertices[1])));
+      for (int e = 0; e < mesh->GetNE(); e++)
+      {
+         if (!el_attr_to_show[mesh->GetAttribute(e) - 1]) { continue; }
+         const auto dx = 0.05 * GetElementLengthScale(e);
+         const auto &ir = fes->GetFE(e)->GetNodes();
+         mesh->GetElementTransformation(e)->Transform(ir, tr);
```
najlkin:
```suggestion
         const Element *el = mesh->GetElement(e);
         const int nv = el->GetNVertices();
         const int *verts = el->GetVertices();
         DenseMatrix &pm = Tr.GetPointMat();
         pm.SetSize(sdim, nv);
         for (int v = 0; v < nv; v++)
         {
            const real_t *vert = mesh->GetVertex(verts[v]);
            for (int d = 0; d < sdim; d++)
            {
               pm(d, v) = vert[d];
            }
         }
         Tr.SetFE(mesh->GetTransformationFEforElementType(el->GetType()));
         Tr.Transform(ir, tr);
```
Unfortunately, this also applies curvilinear mapping if the mesh is nodal. However, I have managed to bypass it ☝️ . (It is essentially user-space version of `Mesh::GetElementTransformation()` for non-nodal mesh)

lib/vssolution.cpp
```diff
@@ -1998,39 +1994,145 @@ void VisualizationSceneSolution::PrepareEdgeNumbering()
...
 
-         ShrinkPoints(p, k, 0, 0);
+   if (shading == Shading::Flat || shading == Shading::Smooth)
+   {
+      H1_FECollection h1_fec(1, mesh->Dimension());
+      FiniteElementSpace h1_fes(mesh, &h1_fec);
+      MFEM_VERIFY(sol && sol->Size() == h1_fes.GetNDofs(),
+                  "Flat space does not match the solution size");
+      GridFunction h1_sol(&h1_fes, sol->GetData());
+      const auto *fes = rsol ? rsol->FESpace() : &h1_fes;
```
najlkin, line 2090:
```suggestion
      const auto *fes = rsol ? rsol->FESpace() : &h1_fes;
      IsoparametricTransformation Tr;
      const int sdim = mesh->SpaceDimension();
```

lib/vssolution.cpp
camierjs: neat, thank you @najlkin !
Inserted in [db99198](https://github.com/GLVis/glvis/pull/339/commits/db99198bab50684dd8d9e2a6edda6959bbc2800d)

lib/vssolution.cpp
camierjs: done in [db99198](https://github.com/GLVis/glvis/pull/339/commits/db99198bab50684dd8d9e2a6edda6959bbc2800d)

lib/vssolution.cpp
```diff
@@ -1825,169 +1833,157 @@ double VisualizationSceneSolution::GetElementLengthScale(int k)
...
+            constexpr double eps = 1e-12;
+            const int rank = mesh->GetAttribute(e) - 1;
+            MFEM_VERIFY(rank >= 0 && rank < (int)offsets->size(),
+                        "Invalid rank");
+            const int l_e = offset(e);
+            const auto &xy = (*offsets)[rank].exy_map.at({l_e,rank});
+            MFEM_VERIFY(fabs(xy.x - xc) < eps && fabs(xy.y - yc) < eps,
+                        "Element center does not match expected position");
+         }
+#endif // GLVIS_DEBUG
```
najlkin, line 1879: This does not work for "parallel" numbering and throws an exception "out_of_range" for `exy_map` 💣 , because `l_e` is not the local index. Either guard this for to perform the check only for local numbering, or override the behavior of `offset`, or ... just delete it 😄

README.md
```diff
@@ -177,9 +177,10 @@ Key commands
 - <kbd>Ctrl</kbd> + <kbd>o</kbd> – Toggle an element ordering curve
 - <kbd>n</kbd> / <kbd>N</kbd> – Cycle through numberings. The options are:
```
najlkin: Mention also Alt+n 😉

camierjs: done in [291cccc](https://github.com/GLVis/glvis/pull/339/commits/291cccc857ce9fba60e946ffa463c0b790d2afaa)

lib/vssolution.cpp
camierjs: Fixed in [4d51a5f](https://github.com/GLVis/glvis/pull/339/commits/4d51a5fa8c5d3a610eae25a17f8277db77d62f2f)

PR #340, opened 2025-05-31
Proper stream parsing in glvis-js

najlkin:
This PR unifies handling of streams in JS and glvis app to enable all stream commands, resolving #315. ~Moreover, it introduces animations, i.e., time series of solutions.~

⭐ bonus:
- Fixed compilation with emscripten on Mac to not use CGL 

↪️ dependencies:

- #336

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ option(GLVIS_USE_EGL
 
 option(GLVIS_USE_CGL
   "Use CGL for headless rendering"
-  (APPLE AND NOT EMSCRIPTEN))
+  APPLE)
 
 #
 # Handle a few other definitions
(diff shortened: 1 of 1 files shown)


najlkin:
To be clear, this PR enables general processing, but only processes a single command after the solution is loaded like it worked so far. It is easily possible to extend this (by calling the new method `communication_thread::process_one()`) and process arbitrary amount of commands, but it requires changes on the side of `glvis-js`. At the moment, we cannot distinguish between files and streams in `aux_js.cpp`, where waiting for the latter would stop the visualization.

najlkin:
This is glvis-js compilation with the latest commit: https://github.com/GLVis/glvis-js/actions/runs/21537548572 (branch `ci-build-21537548572`). Please test if everything works as intended in glvis-js 👈

tzanio:
@najlkin can you resolve conflict?

Also, should this be mentioned in `CHANGELOG`?

tzanio (approved):
I also tested and looks look good, thanks @najlkin !

Issue #341, opened 2025-06-02
f command crashes app


kaylamarietorres:
I am experiencing the f command on GLVIS crashing the app.

najlkin:
Hi, could you elaborate when? Ideally, could you provide an example?

kaylamarietorres:
I was running the Example 15 (Adaptive Mesh Refinement) [./ex15 -n 3] [located here](https://github.com/tzanio/data/blob/main/pisale/README.md), I was practicing the keys to try and get it to look like the example

najlkin:
I could reproduce that also with Example 6. However, it has been/will be resolved by #339 . Can you confirm that PR fixes that for you?

najlkin:
Closing this in favor of #343 .

PR #343, opened 2025-06-03
Fixed resetting of scalar data

najlkin:
Fixed reset of `sol*` vectors for visualization scene, which caused crashes like in #341 .

### DIFF
--- a/lib/data_state.cpp
+++ b/lib/data_state.cpp
@@ -572,7 +572,8 @@ bool DataState::SetNewMeshAndSolution(DataState new_state,
    {
       ResetMeshAndSolution(new_state, vs);
 
-      *this = std::move(new_state);
+      // do not sol vector as it is updated directly
+      internal = std::move(new_state.internal);
 
       return true;
    }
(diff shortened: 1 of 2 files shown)


lib/data_state.cpp
```diff
@@ -572,10 +572,8 @@ bool DataState::SetNewMeshAndSolution(DataState new_state,
    {
       ResetMeshAndSolution(new_state, vs);
 
-      internal.grid_f = std::move(new_state.internal.grid_f);
-      internal.mesh = std::move(new_state.internal.mesh);
-      internal.quad_f = std::move(new_state.internal.quad_f);
-      internal.mesh_quad = std::move(new_state.internal.mesh_quad);
+      // do not sol vector as it is updated directly
```
camierjs: Word missing?

camierjs (approved):
Looks good, thank you @najlkin !

lib/data_state.cpp
najlkin: Fixed 😉

PR #344, opened 2025-06-03
Palette cli

justinlaughlin:
- Adds the option to specify a palette using CLI arg `-pal` or `-palette`
- CLI arg for specifying extra palettes via file is now `-pfile` or `-palette-file`
- Internally, use `PaletteState.SetFallbackIndex()` over `PaletteState.SetIndex()` to specify a fallback index if the default is not specified (implementation would have been more invasive via `PaletteState` so it is done through the global `BasePalettes`)

### DIFF
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -1471,7 +1471,7 @@ int main (int argc, char *argv[])
                   "Run a GLVis script file.");
    args.AddOption(&palette_file, "-pfile", "--palette-file",
                   "Palette file.");
-   args.AddOption(&palette_init, "-pal", "--palette-init",
+   args.AddOption(&palette_init, "-pal", "--palette",
                   "Initial palette.");
    args.AddOption(&arg_keys, "-k", "--keys",
                   "Execute key shortcut commands in the GLVis window.");
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -169,7 +169,7 @@ bool GLVisInitVis(StreamCollection input_streams)
             vs->Zoom(1.8);
             // Use the 'bone' palette when visualizing a 2D mesh only (otherwise
             // the 'jet-like' palette is used in 2D, see vssolution.cpp).
-            vs->palette.SetIndex(4);
+            vs->palette.SetFallbackIndex(4);
          }
       }
       else if (stream_state.mesh->SpaceDimension() == 3)
@@ -186,13 +186,13 @@ bool GLVisInitVis(StreamCollection input_streams)
             if (stream_state.mesh->Dimension() == 3)
             {
                // Use the 'white' palette when visualizing a 3D volume mesh only
-               vss->palette.SetIndex(11);
+               vss->palette.SetFallbackIndex(11);
                vss->SetLightMatIdx(4);
             }
             else
             {
                // Use the 'bone' palette when visualizing a surface mesh only
-               vss->palette.SetIndex(4);
+               vss->palette.SetFallbackIndex(4);
             }
             // Otherwise, the 'vivid' palette is used in 3D see vssolution3d.cpp
             vss->ToggleDrawAxes();
@@ -1417,6 +1417,7 @@ int main (int argc, char *argv[])
    const char *stream_file   = string_none;
    const char *script_file   = string_none;
    const char *palette_file  = string_none;
+   const char *palette_init  = string_none;
    const char *font_name     = string_default;
    int         portnum       = 19916;
    int         multisample   = GetMultisample();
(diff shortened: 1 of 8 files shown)


glvis.cpp
```diff
@@ -1468,8 +1469,10 @@ int main (int argc, char *argv[])
                   "Number of digits used for processor ranks in file names.");
    args.AddOption(&script_file, "-run", "--run-script",
                   "Run a GLVis script file.");
-   args.AddOption(&palette_file, "-pal", "--palettes",
+   args.AddOption(&palette_file, "-pfile", "--palette-file",
```
najlkin: I am not sure about this one. Palettes with `-pal` were already in the release, so we should keep backward compatibility maybe.

najlkin:
How about script and stream commands. Could you add them too? 😉  If you branch this from #335 , it would be very easy. It makes sense especially with palette files, because it is not that easy to sort out what will be the index when loaded to GLVis.

glvis.cpp
justinlaughlin: I think this makes more sense but you're right it breaks compatibility. Do you have any suggestions?

justinlaughlin: `-pname` and `-palette-name`?

najlkin: Yes, for example that 😉

tzanio:
@justinlaughlin, can you resolve the conflicts?

lib/script_controller.cpp
```diff
@@ -722,9 +726,29 @@ void ScriptController::ExecuteScriptCommand()
             int pal;
             scr >> pal;
             cout << "Script: palette: " << pal << endl;
-            win.vs->palette.SetIndex(pal-1);
+            win.vs->palette.SetIndex(pal);
```
najlkin: This breaks backward compatibility, which is not a huge issue as there are just a few users of the scripts, I guess 😁 , but anyway. It should be mentioned in Changelog.

lib/script_controller.cpp
```diff
@@ -722,9 +726,29 @@ void ScriptController::ExecuteScriptCommand()
...
             MyExpose();
          }
+         break;
+         case Command::PaletteFile:
+         {
+            std::string palette_file;
+            scr >> palette_file;
```
najlkin: This will not work with spaces, so detecting a delimiter would be better 🤔 .

lib/script_controller.cpp
justinlaughlin: How does it break backward compatibility? #335 isn't merged yet right? This was just an error from my testing; I think it is because palettes used to be displayed as 1-indexed but now they are 0-indexed.

najlkin: This is wrong 😞 , it is not matching the list of palettes, which starts from 1.

lib/script_controller.cpp
justinlaughlin: Good catch - see [c5de7fd](https://github.com/GLVis/glvis/pull/344/commits/c5de7fd44a45105529ba8dc913352146f5facb8b)

lib/script_controller.cpp
justinlaughlin: image

I thought it was, but at least on this version it starts at 0.

lib/palettes_base.cpp
```diff
@@ -452,7 +469,7 @@ void PaletteRegistry::Load(const string& palette_filename)
          // Add new palette
          AddPalette(palname);
          idx = GetIndexByName(palname);
-         cout << "Reading palette: (" << idx+1 << ") " << palname << endl;
+         cout << "Reading palette: (" << idx << ") " << palname << endl;
```
najlkin: This should be also rather 1-based to match the list, so revert this.

lib/script_controller.cpp
justinlaughlin: .. which was probably caused by me lol. Let me go back and fix.

najlkin: Press F6, the list there starts from 1. I do not know, I do not see a strong reason to change the numbering to 0-based and break backward compatibility 🤔 .

justinlaughlin: Yeah that is just my bad, I got confused 🙃; I incorrectly indexed some print statements last year and thought we changed to 0-based 😆 . See [01b7311](https://github.com/GLVis/glvis/pull/344/commits/01b73116cf27a718acb32985e74a054f4cb8ab3a)

lib/palettes_base.cpp
justinlaughlin: [01b7311](https://github.com/GLVis/glvis/pull/344/commits/01b73116cf27a718acb32985e74a054f4cb8ab3a)

lib/threads.hpp
```diff
@@ -47,7 +47,7 @@ class GLVisCommand
       SHADING = 11,
       VIEW_CENTER = 12,
       AUTOSCALE = 13,
-      PALETTE = 14,
+      PALETTE_FILE = 14,
```
najlkin: Maybe leave the original command here. Besides, this could be refactored, removing the numbers and changing it `enum class`, so the order could be arbitrary 😉 . Do you want to do that?

justinlaughlin: Also a good idea. See 1cec843. Feel free to adjust as you see fit

najlkin: Cool 👍 , you can reorder it now 🚀 . Also please return the `ThreadCommand` where they were, I would prefer them just before `ThreadCommands` as you need to update both when adding something.

justinlaughlin: Done

najlkin (approved):
Looks good! 🎉  Pls mention it in changelog.

glvis.cpp
justinlaughlin: I'll change the script/stream commands to match, but I think I still prefer `--palette-file` over `--palettes`, as the latter is too similar to `--palette`, which is another script/stream command.

PR #345, opened 2025-10-05
CGL headless visualization @ macos

camierjs:
This PR should allow https://github.com/GLVis/glvis/pull/338 to work on MacOS with CGL (Core OpenGL).


lib/egl/egl.hpp
```diff
@@ -115,6 +116,7 @@ class EglWindow : public GLWindow
 
    void signalLoop() override;
 
+   bool initGLEW(bool legacyGlOnly) { return GLWindow::initGLEW(legacyGlOnly); }
```
najlkin:
```suggestion

```

lib/egl/egl.hpp
```diff
@@ -20,16 +21,14 @@
 #endif
 
 #ifdef GLVIS_USE_CGL
-#define GL_SILENCE_DEPRECATION // CGL has been deprecated since MacOS 10.14
-#include <OpenGL/CGLTypes.h>
-#include <OpenGL/CGLCurrent.h>
+#define GL_SILENCE_DEPRECATION
```
najlkin: I would revert this. I think it is nice to have the reason for this declaration. Secondly, CGL is not a part of OpenGL officially, so it is more clear to include it explicitly even when Apple packs it together with OpenGL (also I do not know if in all versions etc.).

lib/egl/egl_main.cpp
```diff
@@ -437,27 +424,24 @@ EglMainThread::Handle EglMainThread::CreateWindow(EglWindow *caller, int w,
...
-   if (!out_hnd.isInitialized())
-   {
-      return out_hnd;
+   if (!out_hnd.isInitialized()) { return out_hnd; }
+   CGLError err = CGLSetCurrentContext(out_hnd.ctx.get());
+   if (err != kCGLNoError) { return out_hnd; }
+   // initialize GLEW before using any OpenGL functions
+   const auto status = caller->initGLEW(legacy_gl);
+   MFEM_VERIFY(status, "GLEW initialization failed!");
+   assert(glGetError() == GL_NO_ERROR);
```
najlkin: `glGetError()` could be checked in `initGLEW()`, so this (and the `#include`) would not be needed 🤔 .

lib/egl/egl_main.cpp
```diff
@@ -437,27 +424,24 @@ EglMainThread::Handle EglMainThread::CreateWindow(EglWindow *caller, int w,
...
-   if (err != kCGLNoError)
-   {
-      std::cerr << "Cannot set the CGL context as current, error: "
-                << CGLErrorString(err) << std::endl;
-
-      return out_hnd;
-   }
-
    glGenFramebuffers(1, &out_hnd.buf_frame);
+   glCheckFramebufferStatus(GL_FRAMEBUFFER);
```
najlkin: Is this good for something? The return value is not checked and the frame buffer is incomplete at this point anyway, because it does not have the render buffers attached yet.

lib/egl/egl_main.cpp
```diff
@@ -437,27 +424,24 @@ EglMainThread::Handle EglMainThread::CreateWindow(EglWindow *caller, int w,
...
+   CGLError err = CGLSetCurrentContext(out_hnd.ctx.get());
+   if (err != kCGLNoError) { return out_hnd; }
+   // initialize GLEW before using any OpenGL functions
+   const auto status = caller->initGLEW(legacy_gl);
+   MFEM_VERIFY(status, "GLEW initialization failed!");
+   assert(glGetError() == GL_NO_ERROR);
+   const auto *renderer = glGetString(GL_RENDERER);
+   if (strstr((const char*)renderer, "Software") != nullptr)
+   {
+      PRINT_DEBUG("Using software renderer; expect limited functionality");
```
najlkin: I am just curious, does this happen often on Mac? 🤔

lib/egl/egl_main.cpp
```diff
@@ -488,16 +472,13 @@ void EglMainThread::ResizeWindow(Handle &handle, int w, int h)
 #ifdef GLVIS_USE_CGL
```
najlkin, line 463: Have you tested resizing an existing window? Does it work on Mac? 🤔

lib/egl/egl.hpp
```diff
@@ -42,9 +41,12 @@ class EglWindow : public GLWindow
       EGLConfig eglCfg{};
 #endif
 #ifdef GLVIS_USE_CGL
-      GLuint buf_frame {}, buf_color {}, buf_depth {};
-      CGLPixelFormatObj pixFmt {};
-      CGLContextObj ctx{};
+      GLuint buf_frame, buf_color, buf_depth;
+      CGLPixelFormatObj pix;
```
najlkin: I think it would not hurt to initialize these to be sure we have a complete handle, but it is true it is implementation dependent 🤔

najlkin (commented):
Thanks @camierjs ! 👍

lib/egl/egl_main.cpp
camierjs: removed in https://github.com/GLVis/glvis/pull/345/commits/e873d0c5c2524b4b1fb6bdc0669970a323cbfd5a

lib/egl/egl.hpp
camierjs: You're right - the warning only shows up with the `pedantic` flags and I brought it back w/o the comment - good catch.

lib/egl/egl_main.cpp
camierjs: never - I removed it in https://github.com/GLVis/glvis/pull/345/commits/e873d0c5c2524b4b1fb6bdc0669970a323cbfd5a

lib/egl/egl_main.cpp
camierjs: yes: it works fine on macos

lib/egl/egl.hpp
camierjs: done in https://github.com/GLVis/glvis/pull/345/commits/e873d0c5c2524b4b1fb6bdc0669970a323cbfd5a

lib/egl/egl.hpp
camierjs: done in https://github.com/GLVis/glvis/pull/345/commits/e873d0c5c2524b4b1fb6bdc0669970a323cbfd5a

lib/egl/egl.hpp
camierjs: done in https://github.com/GLVis/glvis/pull/345/commits/e873d0c5c2524b4b1fb6bdc0669970a323cbfd5a

PR #346, opened 2025-11-25
Support serial mesh coming from each rank in parallel

kmittal2:


najlkin:
Confirmed to be working 👍 , but superseded by #335 .

najlkin (approved):

PR #349, opened 2025-12-05
Cleaned up default palettes

najlkin:
This PR does two things:

- makes `palettes_default.cpp` a regular source file instead of inlining it, which is a bad practice and it is hard to debug
- changes values to `float` literals to avoid the massive amount of warnings during compilation


tzanio (approved):
Thanks @najlkin!

PR #368, opened 2026-03-11
Bug fix for real attributes in parallel socket input (glvis -a)

tzanio:
### Overview

This is a bug fix for `glvis -a` which did not work correctly for parallel socket input.

### Testing

Start the server with 

```
glvis -a
```

then send a solution via socket, from a mesh with multiple attributes, e.g.

```
mpirun -np 4 ./ex1p -m ../data/beam-quad.mesh
```

and check that <kbd>F11</kbd> pulls apart material subdomains, not processor subdomains.

### Potential improvements
- Add the ability to specify this via a socket command?

### DIFF
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -77,14 +77,14 @@ class Session
 
 public:
    Session(bool fix_elem_orient,
-           bool keep_attr,
            bool save_coloring,
+           bool keep_attr,
            string plot_caption,
            bool headless)
    {
       win.data_state.fix_elem_orient = fix_elem_orient;
-      win.data_state.keep_attr = keep_attr;
       win.data_state.save_coloring = save_coloring;
+      win.data_state.keep_attr = keep_attr;
       win.plot_caption = plot_caption;
       win.headless = headless;
    }
@@ -159,7 +159,7 @@ public:
 };
 
 void GLVisServer(int portnum, bool save_stream, bool fix_elem_orient,
-                 bool keep_attr, bool save_coloring, string plot_caption,
+                 bool save_coloring, bool keep_attr, string plot_caption,
                  bool headless = false)
 {
    std::vector<Session> current_sessions;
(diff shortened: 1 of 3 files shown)


glvis.cpp
```diff
@@ -77,11 +77,13 @@ class Session
 
 public:
    Session(bool fix_elem_orient,
+           bool keep_attr,
```
najlkin: Minor, it would be better to unify the order of these parameters with `DataState`, where `keep_attr` goes after `save_coloring`, so I would suggest to either swap it here or there.

najlkin (commented):
Looks good 👍 . It has been broken for a long time and the refactoring fixed that for other places, but even the parameters were missing here, so I was wondering if the bug is not a feature 😅.
We have a script command for that (`psolution`), but stream commands work differently and each part is sent with just `solution`, so we cannot just add easily a parameter there. Perhaps we could add some commands which would go before solution? 🤔 But we do not have that for other things and it requires to change how things are constructed and processed. Right now, `StreamReader` reads the stream and initializes `DataState` from which the visualization is initialized and only then processing is passed to `communication_thread`, which processes the rest. If we do not want change this completely, we could think of adding these initialization commands to `StreamReader` and let it process as many as needed until it reaches some solution command 🤔

najlkin:
The changes for stream commands for keeping attributes and other things are in #369 .

glvis.cpp
tzanio: Fixed in 466bc477a63025b1b8a1a20087813d86840db302

dylan-copeland (approved):
Thanks @tzanio! I tested, and everything looks fine. My only suggestion is to improve the documentation for the implementation of this functionality. It looks like `keep_attr` controls whether `DataState::SetGridFunction` calls `ComputeDofsOffsets`, and it is not clear to me why this would determine whether `F11` splits the domain by attribute vs. parallel partition. Documentation could be added for `keep_attr` and these functions, as well as how they determine the behavior of `F11`.

najlkin:
@dylan-copeland , `ComputeDofsOffsets` is newly used (#339 ) for DOFs numbering. It depends on the attributes just coincidentally in the sense that the implementation needs to know from what rank the DOFs are coming from, where @camierjs used the attribute when it is not a real one but the MPI rank. It is a detail, but it is true it could be mentioned somewhere in documentation.
That `F11` pulls apart subdomains of different attributes whatever they represent. README is a bit misleading here, because it says "material subdomains", which is not always the case. This could be clarified.

camierjs (approved):

PR #373, opened 2026-07-16
[BUG] Fixed (and refactored) geometry refiner

najlkin:
Fixed geometry refiner settings, which was not correctly passed to the worker threads as reported by @v-dobrev . This was to great extent a problem of the poor design of the shared global object. This PR refactors the codebase, moving the `GeometryRefiner` to the base class `VisualizationSceneScalarData`, where it is available to all visualization scenes. The type of the refiner is set in `DataState` along other settings. Moreover, passing of this kind of parameters has been refactored out from `GLVisServer` as the list was growing and growing 😳 .

### DIFF
--- a/glvis.cpp
+++ b/glvis.cpp
@@ -75,13 +75,6 @@ class Session
    std::thread handler;
 
 public:
-   Session(string plot_caption,
-           bool headless)
-   {
-      win.plot_caption = plot_caption;
-      win.headless = headless;
-   }
-
    Session(Window other_win)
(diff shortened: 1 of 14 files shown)


tzanio (approved):
Thanks for fixing this, @najlkin !

camierjs (approved):
Looks great: thank you @najlkin !

najlkin:
@v-dobrev , does the patch work for you? Should we merge?

PR #374, opened 2026-07-16
Add direct cutting-plane control to GLVis scripts

arotem3:
## Motivation
Today, the only way to position the 3D cutting plane from a GLVis script is to replay raw keystrokes via keys. Even a modest plane orientation requires several keystrokes:
```
keys xxxxxyyyyyyyyi
```
(5 `x` presses to rotate `phi` by 25°, 8 `y` presses to rotate `theta` by 40°, then `i` to enable the plane). Each individual keystroke triggers a full cutting-plane geometry recomputation, not just the final one. On a large mesh that's a lot of wasted computation just to reach a state that could be stated directly in one line.

## What this adds
A single script command:
```
cutting_plane <phi> <theta> <translation> <kind:optional> <alg:optional>
```
which sets the plane's orientation (degrees), translation, kind (cut-through-elements / solid-cut-behind / off), and algorithm (accurate / fast) directly, in one shot reproducing exactly what the equivalent keystroke sequence would have produced, without the redundant intermediate recomputations. `kind` defaults to 1 (cut through elements) so the plane is visible immediately; `alg` is optional and leaves the current value unchanged when omitted.

## Implementation
- `Plane::SetPlane` : new direct setter on the existing `Plane` class (previously only relative `Increase/Decrease` mutators existed).
- `VisualizationSceneScalarData::SetCuttingPlane` : empty-default virtual hook (2D scenes no-op safely).
`VisualizationSceneSolution3d::SetCuttingPlane` — override combining plane move + kind + algorithm changes, mirroring the existing key-handler logic.
- `cutting_plane` registered in the script command table.

### DIFF
--- a/lib/script_controller.cpp
+++ b/lib/script_controller.cpp
@@ -131,7 +131,7 @@ ScriptCommands::ScriptCommands()
    (*this)[Command::PlotCaption]          = {"plot_caption", "'<caption>'", "Set the plot caption."};
    (*this)[Command::PointLine]            = {"pointline", "<num_points> <x y z>...", "Set point line overlay coordinates."};
    (*this)[Command::Headless]             = {"headless", "", "Change the session to headless."};
-   (*this)[Command::CuttingPlane]         = {"cutting_plane", "<phi> <theta> <translation> <kind:optional, default 1> <alg:optional>", "Set the cutting plane orientation (degrees), translation, kind, and algorithm."};
+   (*this)[Command::CuttingPlane]         = {"cutting_plane", "<phi> <theta> <translation> [<kind> [<alg>]]", "Set the cutting plane orientation (degrees), translation, kind (default: 1), and algorithm."};
 }
 
 int ScriptController::ScriptReadSolution(istream &scr, DataState &state)
--- a/lib/script_controller.cpp
+++ b/lib/script_controller.cpp
@@ -871,12 +871,12 @@ bool ScriptController::ExecuteScriptCommand()
             int kind = 1, algo = -1;
             scr >> ws;
             int ch = scr.peek();
-            if (ch != std::char_traits<char>::eof() && (isdigit(ch) || ch == '-'))
+            if (isdigit(ch) || ch == '-')
             {
                scr >> kind;
                scr >> ws;
                ch = scr.peek();
-               if (ch != std::char_traits<char>::eof() && (isdigit(ch) || ch == '-'))
+               if (isdigit(ch) || ch == '-')
                {
                   scr >> algo;
                }
--- a/lib/script_controller.cpp
+++ b/lib/script_controller.cpp
@@ -870,14 +870,29 @@ bool ScriptController::ExecuteScriptCommand()
 
             int kind = 1, algo = -1;
             scr >> ws;
-            if (isdigit((unsigned char)scr.peek()) || scr.peek() == '-')
+            int ch = scr.peek();
+            if (ch != std::char_traits<char>::eof() && (isdigit(ch) || ch == '-'))
             {
                scr >> kind;
                scr >> ws;
-               if (isdigit((unsigned char)scr.peek()) || scr.peek() == '-')
+               ch = scr.peek();
+               if (ch != std::char_traits<char>::eof() && (isdigit(ch) || ch == '-'))
                {
                   scr >> algo;
                }
+            }
+
+            if (kind < -1 || kind > 2)
+            {
+               cerr << "Script: cutting_plane: invalid kind " << kind
+                    << " (expected -1..2)" << endl;
+               kind = 1;
+            }
+            if (algo != -1 && (algo < 0 || algo > 1))
+            {
+               cerr << "Script: cutting_plane: invalid alg " << algo
+                    << " (expected 0 or 1)" << endl;
+               algo = -1;
             }
 
             cout << "Script: cutting_plane: " << phi_deg << ' ' << theta_deg
--- a/lib/script_controller.cpp
+++ b/lib/script_controller.cpp
@@ -888,10 +888,10 @@ bool ScriptController::ExecuteScriptCommand()
                     << " (expected -1..2)" << endl;
                kind = 1;
             }
-            if (algo != -1 && (algo < 0 || algo > 1))
+            if (algo < -1 || algo > 1)
             {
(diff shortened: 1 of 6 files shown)


lib/script_controller.cpp
```diff
@@ -859,6 +863,49 @@ bool ScriptController::ExecuteScriptCommand()
...
             break;
+         case Command::CuttingPlane:
+         {
+            double phi_deg, theta_deg, translation;
+            scr >> phi_deg >> theta_deg >> translation;
+
+            int kind = 1, algo = -1;
+            scr >> ws;
+            int ch = scr.peek();
+            if (ch != std::char_traits<char>::eof() && (isdigit(ch) || ch == '-'))
```
najlkin: Remove the EOF check, it does not make sense - EOF is not a digit nor minus sign 🤪 .

lib/vssolution3d.cpp
```diff
@@ -1192,6 +1192,36 @@ void VisualizationSceneSolution3d::ToggleCPAlgorithm()
...
+      cp_algo = algo;
+   }
+
+   CPPrepare();
+   if (cplane == 0 || cplane == 2)
+   {
+      Prepare();
+      PrepareLines();
+      PrepareOrderingCurve();
+   }
```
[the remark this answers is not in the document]
najlkin: Well, it is right, but the suggestion is unnecessarily convoluted:

- `cplane_old` makes sense to save
- `cplane_changed` does *not* make sense, it is trivial to compare `cplane` and `cplane_old`
- the test condition is correct `if (cplane_old == 2 || cplane == 2)`
- the inner test for the ordering curve makes sense to do, but it should be only `if (cplane != cplane_old)`
- the second branch does *not* make sense, because it does not realize that it was just the leaving case in the toggling procedure as ` (2+1) % 3 == 0` 🙃 .

lib/script_controller.cpp
arotem3: Done. See [4a22f38](https://github.com/GLVis/glvis/pull/374/commits/4a22f38b29804ccc8c7da590e7b5417a9b732c9b)

lib/vssolution3d.cpp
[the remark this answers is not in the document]
arotem3: Fix attempted at [cc8b16a](https://github.com/GLVis/glvis/pull/374/commits/cc8b16ae6a657453f530860ac8555d1505ef65bb)

lib/vsdata.hpp
```diff
@@ -263,6 +265,12 @@ class VisualizationSceneScalarData : public VisualizationScene
    virtual void AutoRefine() = 0;
    virtual void ToggleAttributes(mfem::Array<int> &attr_list) = 0;
 
+   // Set the cutting-plane orientation (radians), translation, kind, and
+   // algorithm in one shot. No-op by default; only meaningful for 3D scenes.
+   // kind == -1 / algo == -1 mean "leave unchanged".
```
najlkin: Please use doxygen style. We do not publish it anymore on the webpage, but you can still build it locally.

lib/vsdata.hpp
```diff
@@ -263,6 +265,12 @@ class VisualizationSceneScalarData : public VisualizationScene
...
+   virtual void SetCuttingPlane(double phi, double theta, double translation,
+                                int kind, int algo) { }
```
najlkin:
```suggestion
   virtual void SetCuttingPlane(double phi, double theta, double translation,
                                int kind = -1, int algo = -1) { }
```

lib/vssolution3d.hpp
```diff
@@ -218,6 +218,8 @@ class VisualizationSceneSolution3d : public VisualizationSceneScalarData
    void ToggleCPDrawElems();
    void ToggleCPDrawMesh();
    void ToggleCPAlgorithm();
+   void SetCuttingPlane(double phi, double theta, double translation,
+                        int kind, int algo) override;
```
najlkin, line 222:
```suggestion
   void SetCuttingPlane(double phi, double theta, double translation,
                        int kind = -1, int algo = -1) override;
```

lib/script_controller.cpp
```diff
@@ -859,6 +863,49 @@ bool ScriptController::ExecuteScriptCommand()
...
+               }
+            }
+
+            if (kind < -1 || kind > 2)
+            {
+               cerr << "Script: cutting_plane: invalid kind " << kind
+                    << " (expected -1..2)" << endl;
+               kind = 1;
+            }
+            if (algo != -1 && (algo < 0 || algo > 1))
```
najlkin:
```suggestion
            if (algo < -1 || algo > 1)
```

lib/script_controller.cpp
```diff
@@ -859,6 +863,49 @@ bool ScriptController::ExecuteScriptCommand()
...
+            {
+               cerr << "Script: cutting_plane: invalid alg " << algo
+                    << " (expected 0 or 1)" << endl;
```
najlkin: Fix this message please to include -1.

lib/script_controller.cpp
```diff
@@ -128,6 +131,7 @@ ScriptCommands::ScriptCommands()
    (*this)[Command::PlotCaption]          = {"plot_caption", "'<caption>'", "Set the plot caption."};
    (*this)[Command::PointLine]            = {"pointline", "<num_points> <x y z>...", "Set point line overlay coordinates."};
    (*this)[Command::Headless]             = {"headless", "", "Change the session to headless."};
+   (*this)[Command::CuttingPlane]         = {"cutting_plane", "<phi> <theta> <translation> <kind:optional, default 1> <alg:optional>", "Set the cutting plane orientation (degrees), translation, kind, and algorithm."};
```
najlkin:
```suggestion
   (*this)[Command::CuttingPlane]         = {"cutting_plane", "<phi> <theta> <translation> [<kind> [<alg>]]", "Set the cutting plane orientation (degrees), translation, kind (default: 1), and algorithm."};
```

tzanio (approved):
That's very cool, thanks @arotem3!

najlkin (approved):
LGTM 🎉 . Thanks, @arotem3 👍

camierjs (approved):
Looks good to me!
Thank you @arotem3 for this contribution!

